PR #1, opened 2016-04-19
dictionary module added

AChelikani:
Uses top definition from Glosbe API

### DIFF
--- a/modules/src/__init__.py
+++ b/modules/src/__init__.py
@@ -1,5 +1,6 @@
 __all__ = [
     'movie',
     'joke',
-    'hello'
+    'hello',
+    'dictionary'
 ]
--- a/modules/src/__init__.py
+++ b/modules/src/__init__.py
@@ -1,6 +1,6 @@
 __all__ = [
     'movie',
+    'dictionary',
     'joke',
-    'hello',
-    'dictionary'
+    'hello'
 ]
--- a/modules/src/dictionary.py
+++ b/modules/src/dictionary.py
@@ -1,0 +1,18 @@
+import re
+import requests
+
+def match(input):
+	return bool(re.match(r'^.*\s+definition$', input))
+
+def process(input):
+	output = {}
+	word = re.match(r'^(?P<definition>.*)\s+definition$', input).group('definition')
+	try:
+		r = requests.get("https://glosbe.com/gapi/translate?from=eng&dest=eng&format=json&phrase=" + word + "&pretty=true")
+		data = r.json()
+		output['input'] = input
+		output['output'] = "Definition: " + data['tuc'][0]['meanings'][1]['text']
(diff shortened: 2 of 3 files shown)


modules/src/__init__.py
```diff
@@ -1,5 +1,6 @@
 __all__ = [
     'movie',
     'joke',
-    'hello'
+    'hello',
+    'dictionary'
```
swapagarwal: Since this list is processed in order, I'd like the `dictionary` module to be just after the `movie` one.

modules/src/dictionary.py
```diff
@@ -0,0 +1,18 @@
...
+	return bool(re.match(r'^.*\s+definition$', input))
+
+def process(input):
+	output = {}
+	word = re.match(r'^(?P<definition>.*)\s+definition$', input).group('definition')
+	try:
+		r = requests.get("https://glosbe.com/gapi/translate?from=eng&dest=eng&format=json&phrase=" + word + "&pretty=true")
+		data = r.json()
+		output['input'] = input
+		output['output'] = "Definition: " + data['tuc'][0]['meanings'][1]['text']
```
swapagarwal: Why the index 1?

swapagarwal:
Hey! Nice work, this should be good to merge. Just take a look at the lines that I have commented on. That should do it. :smile:

modules/src/dictionary.py
AChelikani: The response from the API is ordered so that the index 0 has info about the word and language, while the definitions start from index 1.

swapagarwal: Can you check again? See this query `https://glosbe.com/gapi/translate?from=eng&dest=eng&format=json&phrase=cloud&pretty=true` and the output you're giving.
Also, it'd be better if we include the word also in our output.

AChelikani: Hmm, for cloud the 1 index gives the second definition, but for some words like 'flotsam', the 0 index simply just returns the word back. It seems as though the 1 index will always give some viable definition. Any suggestions though?

AChelikani: I've added the input word to the output though.

swapagarwal: Well, `flotsam` is definitely an edge case though.
Some words return only one meaning like `https://glosbe.com/gapi/translate?from=eng&dest=eng&format=json&phrase=ropeway&pretty=true` so I'd like to stick to 0-index.

swapagarwal:
Well, flotsam is definitely an edge case though.
Some words return only one meaning like https://glosbe.com/gapi/translate?from=eng&dest=eng&format=json&phrase=ropeway&pretty=true so I'd like to stick to 0-index.

swapagarwal:
Thanks @AChelikani! :+1:

PR #9, opened 2016-04-22
Wiki module, another type of jokes, minor edits

ymittal:
I added a wiki module but [Wit.ai](https://wit.ai/swapagarwal/JARVIS-on-Messenger) needs some training data. You would need to add a `wiki` Intent and a `query` entity. To verify that my code works without even tweaking Wit.ai, you may hardcode intent as `movie` and visit `http://localhost:5000/process/?q=movie%20happyness`. You should see Wikipedia results for [Happiness](https://en.wikipedia.org/wiki/Happiness).

I also added a different variety of jokes to `joke.py`. I felt our community would appreciate diversity. I tried to look for APIs to get a random joke, but I could not find any free ones.

Thank you!


swapagarwal:
Thanks so much for taking the time to contribute!
I'd really appreciate it if you could open another Pull Request for the `joke` module. (Stick to one module per PR :smile: ) The jokes definitely need to be as diverse as possible, and two APIs are better than one! :stuck_out_tongue_winking_eye:

The `wiki` module looks great! Just take a look at the lines that I have commented on. I have added the intent `wiki` so can you add some tests in the corresponding folder?

modules/__init__.py
```diff
@@ -31,4 +31,4 @@ def search(input):
         else:
             return 'Something didn\'t work as expected! I\'ll report this to my master.'
     else:
-        return 'I\'m sorry; I\'m not sure I understand what you\'re trying to say sir.\nTry typing "help" or "request"'
+        return 'I\'m sorry; I\'m not sure I understand what you\'re trying to say sir.\nTry typing "help" or "request"'
```
swapagarwal: Touch as few files as needed.

modules/src/__init__.py
```diff
@@ -4,5 +4,6 @@
     'help',
     'joke',
     'movie',
+    'wiki',
```
swapagarwal: Alphabetical order.

modules/src/joke.py
```diff
@@ -1,13 +1,17 @@
...
-        r = requests.get('https://api.chucknorris.io/jokes/random')
-        data = r.json()
         output['input'] = input
-        output['output'] = data['value']
+        if randint(0,1):
+            data = requests.get('https://api.chucknorris.io/jokes/random').json()
+            output['output'] = data['value']
+        else:
+            data = requests.get('http://api.yomomma.info/').json()
+            output['output'] = data['joke']
```
swapagarwal: Make this same for now, these changes can be made in another PR.

modules/src/movie.py
```diff
@@ -12,4 +12,4 @@ def process(input, entities):
         output['success'] = True
     except:
         output['success'] = False
-    return output
+    return output
```
swapagarwal: Touch as few files as possible.

modules/src/request.py
```diff
@@ -1,8 +1,8 @@
 def process(input, entities=None):
     request = '''
-Kindly head over to https://github.com/swapagarwal/JARVIS-on-Messenger/issues/new to file an issue:\n
-  - If you find a bug (the bot couldn't handle the query and / or gave irrelevant results), include your search query and the expected result.\n
-  - If you'd like to request a new functionality, feel free to do so by including some sample queries and their corresponding results.\n
+    Kindly head over to https://github.com/swapagarwal/JARVIS-on-Messenger/issues/new to file an issue:\n
+    - If you find a bug (the bot couldn't handle the query and / or gave irrelevant results), include your search query and the expected result.\n
+    - If you'd like to request a new functionality, feel free to do so by including some sample queries and their corresponding results.\n
```
swapagarwal: We want to make this PR about wiki only (so that someone can refer it in the future for adding another module). Make this same for this PR.

modules/src/wiki.py
```diff
@@ -0,0 +1,17 @@
+import sys
+import os
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..', '..', 'lib'))
+import wikipedia
+
+def process(input, entities):
+    output = {}
+    try:
+        query = entities['query'][0]['value']
```
swapagarwal: entities['wikipedia_search_query'][0]['value']

requirements.txt
```diff
@@ -16,3 +16,4 @@ pytest==2.9.1
 requests==2.9.1
 six==1.10.0
 Werkzeug==0.11.8
+beautifulsoup4
```
swapagarwal: version?
Use `pip freeze > requirements.txt`

modules/src/joke.py
havanagrawal: Also, some of these jokes are pretty offensive (being yo momma jokes), I wonder if it's appropriate to have it on a chat bot... (We wouldn't want a [Tay ](https://en.wikipedia.org/wiki/Tay_(bot))fiasco :) )

swapagarwal: Agreed, that's why I'm pushing for a limited set of hand-picked approved jokes.

requirements.txt
ymittal: I wonder if `wikipedia==1.4.0` would install `beautifulsoup4` automatically.

modules/src/joke.py
ymittal: @havanagrawal, that's a good point.

requirements.txt
swapagarwal: Yes, bs4 is a required package in wikipedia (`pip show wikipedia`).
But let's say some other module uses bs4 (for which we're dependent on wikipedia to install) and then we decide, for some reason, to remove wikipedia!
So, to avoid breaking things, it's better to be explicit and add `beautifulsoup4==4.4.1` :smile:

ymittal: That would be an easy fix on the first build. I feel adding `wikipedia==1.4.0` would be more explicit here because bs4 is going to be installed anyway. If any upcoming module uses bs4 and `wikipedia` is removed, we can simply add `beautifulsoup4==4.4.1`. Thoughts?

ymittal:
Thank you for your comments, @swapagarwal. I added `lib/wikipedia` to make sure that `import wikipedia` does not throw an error. `wiki.py` is not in root directory which is why I got an `ImportError` without the folder.

requirements.txt
swapagarwal: Though it is debatable, keeping all the indirect dependencies in requirements.txt with their version numbers makes sure the builds are repeatable. If a module is dependent on five others, and one of them gets an upgrade in a way that is not backward-compatible, that'd be a problem for us! That's why versioning is important.

swapagarwal:
Run `pip install wikipedia` 
That way, the lib files won't be required, and you can simply use `import wikipedia`

requirements.txt
ymittal: Point taken. Adding just `beautifulsoup4==4.4.1`.

ymittal:
I did `pip install wikipedia`. For some reason, I was receiving an `ImportError` until now. I have removed `lib/wikipedia` folder (yay).

swapagarwal:
Are we ready to merge and deploy?

ymittal:
Sure! :smile:

PR #11, opened 2016-04-22
Add book module

havanagrawal:
This module is identical in structure and working to all the other ones.. Basically makes a call to the Goodreads API (You can sign up for a Goodreads Developer Key [here](https://www.goodreads.com/api)

Small difference is that the api does not offer json output, so I've used some minor xml parsing to get the data.

Updates intents and entities on https://wit.ai/havanagrawal/JARVIS-on-Messenger

Sample output
book_usage


config.py
```diff
@@ -1,3 +1,4 @@
 ACCESS_TOKEN = '<ACCESS_TOKEN>'
 VERIFY_TOKEN = '<VERIFY_TOKEN>'
 WIT_AI_ACCESS_TOKEN = 'IKJJJYYVR3X672DHFVS7U7C4L2MQSS2P'
+GOODREADS_ACCESS_TOKEN = '<GOODREAD_ACCESS_TOKEN>'
```
swapagarwal: *GOODREADS

modules/src/book.py
```diff
@@ -0,0 +1,29 @@
+import re
```
swapagarwal: unused

modules/src/book.py
```diff
@@ -0,0 +1,29 @@
...
+        book_node = data.find('book')
+        title = book_node.find('title').text
+        description = book_node.find('description').text
+        link = book_node.find('link').text
+        average_rating = book_node.find('average_rating').text
+        
+        output_data = (title, description, average_rating, link)
+        
+        output['input'] = input
+        output['output'] = "Title: %s<br/>Description: %s<br/>Average Rating: %s<br/> Book link: %s<br/>" % output_data
```
swapagarwal: We need to attribute the data to [Goodreads](https://www.goodreads.com/api/terms). Include something like Powered by Goodreads.

swapagarwal:
That is one awesome module you built there!
I've set up both goodreads and wit on my side. Can you add some tests so that we can see it in action?

modules/src/book.py
havanagrawal: According to the API Guidelines, I was supposed to provide a link back to the original content, which I did using the link output. Should I add "Powered by Goodreads" too?

swapagarwal: Refer the second point in the ToS.

havanagrawal:
Anything else I need to do here?

swapagarwal:
Nope, looks perfect! (Sorry, I got caught up fixing a bug so I didn't see this earlier.)

swapagarwal:
Deployed, and working as expected. Try it out with "anything you want book".
Thanks @havanagrawal! :+1:

PR #12, opened 2016-04-22
Add another API for jokes

havanagrawal:
I wonder if, instead of developing our own list of jokes, why not look for repos on GitHub that might have aggregated such a list?
Perhaps contributing to, and then somehow sourcing from such repos would prove to be more productive.. Just a thought..

API Credit: https://github.com/KiaFathi/tambalAPI

Some other possible sources:
https://github.com/CallumBarclay/Jokes-API
https://github.com/ChrisMcKenzie/dadjokes

### DIFF
--- a/modules/src/joke.py
+++ b/modules/src/joke.py
@@ -1,13 +1,21 @@
 import re
 import requests
+import random
 
 def process(input, entities=None):
     output = {}
+    joke_api_list = ['http://tambal.azurewebsites.net/joke/random',
+                     'https://api.chucknorris.io/jokes/random']
+    joke_key_list = ['joke', 'value']
+    api_count = len(joke_api_list)
+
     try:
-        r = requests.get('https://api.chucknorris.io/jokes/random')
+        index = random.randint(0, api_count - 1)
+        r = requests.get(joke_api_list[index])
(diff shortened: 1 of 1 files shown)


antixcode6:
I agree that this is a good idea - maybe using a small web scraping script to pull from parts of the web with an abundance of jokes. 

Ex. https://www.reddit.com/r/dadjokes

Then grab the title / body of the posts and push it to the other projects

swapagarwal:
@havanagrawal Great finds!
This API is much better than the current one, so how about we replace it (instead of keeping both) ?
And yes, we shouldn't be reinventing the wheel when we can devote that time to expand to support another type of query. :smile:

havanagrawal:
I feel we should keep both, as they both are different genres (one-liners vs Chuck Norris).. Eventually, we can take genre requests too for the joke api, and these will be handy then.. But by all means, if you want me to replace it, I'll replace it :smile:

swapagarwal:
I also want to keep both but if you query the chuck norris api multiple times, you'll find some offensive jokes that I want to avoid. We have users requesting for different genre jokes so that's our next priority. But for now, our main aim is to filter out jokes (which the tambal api already has).

PR #18, opened 2016-04-23
Currency feature

havanagrawal:
Using the API from fixer.io

Updated intents and entities on https://wit.ai/havanagrawal/JARVIS-on-Messenger


swapagarwal:
Whoa, you sure are fast!
The API looks nice, however wouldn't it be better if we could support conversion instead of just supplying the rate? (like 100 USD to INR)

havanagrawal:
Thanks :smile: 
I think the test is failing because I tested locally with my own Wit access token, but didn't commit the token. So it tries to run the test, but with the original token, it has no clue about "currency" as an intent (https://wit.ai/swapagarwal/JARVIS-on-Messenger/eval?_t=83&q=convert+USD+to+HKD)?

Could you please try adding the intent "currency", and the entities "from_currency" and "to_currency", like I've set up on https://wit.ai/havanagrawal/JARVIS-on-Messenger ? If the test still fails, I'll make it more robust.

havanagrawal:
Also, I'll try to get Wit to recognize the amount along with the currencies, i.e. completely support conversion.

swapagarwal:
Ofcourse, my bad! I'll set it up and restart the build.

havanagrawal:
I've used `wit/number` for the amount, since using `wit/amount_of_money` was also grabbing the unit, like usd and inr, along with the number.

Again, the tests will fail, so can you please update your side of Wit?
https://wit.ai/havanagrawal/JARVIS-on-Messenger/intents/currency

swapagarwal:
I restarted the build and all tests passed.
I didn't have to update wit as the tests don't capture the new feature (amount).
For date, it's better to keep it as is so that we know it works! (the api may change anytime)
Is there a collaboration feature in wit?

havanagrawal:
Cool.. I've dropped Wit Help a message.. Hopefully they'll reply within the week.. Anything else I need to do here?

swapagarwal:
Update the tests to include `from`, `to`, and `amount` as well.

havanagrawal:
I've added 2 tests for `from_currency`, `to_currency` and `number`
Could you please (when you get time) add some training data on your side so the tests pass?

You can pick them up from https://wit.ai/havanagrawal/JARVIS-on-Messenger/intents/currency

swapagarwal:
Why delete movie.py?

swapagarwal:
Thanks @havanagrawal! :+1:

PR #20, opened 2016-04-23
Wikipedia URL functionality 

AChelikani:
Added URL for wikipedia page


modules/src/wiki.py
```diff
@@ -4,9 +4,10 @@ def process(input, entities):
     output = {}
     try:
         query = entities['wikipedia_search_query'][0]['value']
+        page = wikipedia.page(query)
         data = wikipedia.summary(query, sentences=1)
```
swapagarwal: We're making two queries here!
Read the docs [here](https://wikipedia.readthedocs.org/en/latest/code.html#module-wikipedia) to figure out how to do the same using only one query. (Hint: see the return type of `wikipedia.page`)

AChelikani: We can do it with one call by doing data=page.summary but then it becomes hard to only give the user the first sentence of the summary, so I thought making two calls would be fine. Let me know.

swapagarwal: I've made some refactoring so if you could fork and send a PR again, that'd be great!
Making fewer calls is our top priority. Take the full summary and wrap it like [this](https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/modules/src/movie.py), setting url in post_text.

swapagarwal:
Hey @AChelikani, will you be willing to complete this?

AChelikani:
@swapagarwal Definitely! Sorry, I've been busy - I'll get to it before the end of the week.

AChelikani:
Fixed! Let me know if it looks good!

modules/src/wiki.py
```diff
@@ -1,14 +1,20 @@
...
 from templates.text import TextTemplate
 
 def process(input, entities):
     output = {}
     try:
         query = entities['wikipedia_search_query'][0]['value']
-        data = wikipedia.summary(query)
+        data = wikipedia.page(query)
+        template = TextTemplate()
+        printable = set(string.printable)
```
swapagarwal: no need of set

modules/src/wiki.py
```diff
@@ -1,14 +1,20 @@
...
+        s = filter(lambda x : x in printable, data.summary)
```
swapagarwal: do we need to do this? can you provide an example why this is required?

modules/src/wiki.py
```diff
@@ -1,14 +1,20 @@
...
+        template.set_text(s)
+        template.set_post_text(data.url)
         output['input'] = input
-        output['output'] = TextTemplate('Wikipedia Summary: ' + data).get_message()
+        output['output'] = template.get_message()["text"]
```
swapagarwal: ["text"] should not be there

modules/src/wiki.py
```diff
@@ -1,14 +1,20 @@
...
         output['success'] = True
     except:
         output['success'] = False
-    return output
+    return output
```
swapagarwal: newline at the end of file (for keeping diffs short)

swapagarwal:
@AChelikani no problem, take your time. just a couple of small fixes and this should be ready to merge.

modules/src/wiki.py
AChelikani: screen shot 2016-05-03 at 11 18 11 am

This is what happens when we don't remove non-printable characters and don't get the ["text"] attribute on get_message

AChelikani: Let me know if I should leave it as is or if there is a good fix for this.

swapagarwal: The summary seems to be rendering correctly, so we can remove the filtering of string.printable.
And the ["text"] is not required; I had made some changes so that get_message() returns the actual message sent to the Messenger API (it includes the 'text' in the json)

AChelikani:
Fixed!

swapagarwal:
Thanks @AChelikani! :+1:

PR #21, opened 2016-04-24
Module for pulling random quotes

AChelikani:
Sample queries:

generate quote, quote, quote please, can i have a quote, random quote

Right now uses a API with only programming/CS related quotes.


swapagarwal:
Fastest new module I ever merged! :smile: 

Thanks @AChelikani! :+1:

PR #24, opened 2016-04-24
Add a new feature prototype

wallid:
I added a feature that scrapped off reddit news from a  country.

### DIFF
--- a/modules/src/news.py
+++ b/modules/src/news.py
@@ -1,0 +1,22 @@
+import re
+import requests
+
+# Also not sure exactly what entities is?
+# Used as their are often some ads displayed on the first few headlines.
+HEADLINE_NUM = 3
+
+def process(input, entities):
+	output = {}
+	try :
+		country = entities['country'][0]['value']
+		country = "Australia"
+		r = requests.get("https://www.reddit.com/r/"+ country.capitalize())
+		news_data = r.json()
+		output['input'] = input
+		output['output'] = 'Headline news in ' + country.capitalize()  + '\n' + news_data['data']['children'][HEADLINE_NUM]['data']['title']
+		output['success']  = True
+	except:
+		output['success'] = False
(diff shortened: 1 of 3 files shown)


swapagarwal:
I've made some refactoring so if you could fork and send a PR again, that'd be great!
Is there an API that we could use for showing news?

modules/src/news.py
```diff
@@ -0,0 +1,22 @@
...
+		r = requests.get("https://www.reddit.com/r/"+ country.capitalize())
+		news_data = r.json()
+		output['input'] = input
+		output['output'] = 'Headline news in ' + country.capitalize()  + '\n' + news_data['data']['children'][HEADLINE_NUM]['data']['title']
+		output['success']  = True
+	except:
+		output['success'] = False
+	return output
+
+process(input)
```
swapagarwal: no need of this line, it's already being called in the `search` function [here](https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/modules/__init__.py)

modules/src/news.py
```diff
@@ -0,0 +1,22 @@
+import re
+import requests
+
+# Also not sure exactly what entities is?
```
swapagarwal: Entities will return relevant text from the query, try `define a superhero` in try out a sentence at the top [here](https://wit.ai/swapagarwal/JARVIS-on-Messenger).

modules/src/news.py
```diff
@@ -0,0 +1,22 @@
...
+# Used as their are often some ads displayed on the first few headlines.
+HEADLINE_NUM = 3
```
swapagarwal: This is a bit hacky! Can you search for an alternative? An API, maybe?

swapagarwal:
Hey @wallid, will you be willing to complete this? :smile:

PR #30, opened 2016-04-29
added ping module

sasek333:
requires os's ping access

### DIFF
--- a/modules/src/ping.py
+++ b/modules/src/ping.py
@@ -1,0 +1,26 @@
+import subprocess
+import platform
+
+def process(input, entities=None):
+
+    output = {'input': input}
+
+    host = entities['host'][0]['value']
+
+    if not host:
+        output['success'] = False
+        output['output'] = "Please specify a host"
+
+        return output
+
+    ping_str = "-n 1" if platform.system().lower() == "windows" else "-c 1"
+    args = "ping " + " " + ping_str + " " + host
+    need_sh = False if platform.system().lower() == "windows" else True
+
+    text =  'System reachable' if subprocess.call(args, shell=need_sh) == 0 else 'System unrechable'
+
+    output['output'] = test
+    output['success'] = True
+
+    return output
+
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,6 @@ target/
 
 #Ipython Notebook
 .ipynb_checkpoints
+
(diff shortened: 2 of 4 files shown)


swapagarwal:
Can we leverage an API call for this?

sasek333:
You mean like pyping? It requires root on linux machines. I'll check it in a week because from today it's vacation time :)

modules/src/ping.py
```diff
@@ -0,0 +1,26 @@
...
+
+    host = entities['host'][0]['value']
+
+    if not host:
+        output['success'] = False
+        output['output'] = "Please specify a host"
+
+        return output
+
+    ping_str = "-n 1" if platform.system().lower() == "windows" else "-c 1"
```
swapagarwal: this `platform` thing is bugging me!

swapagarwal:
Keeping in mind that it should be cross-platform compatible, we have two options: either use a python library (sockets / requests) to check website availability, or use an external service like [`Is It Up`](https://isitup.org/github.com).
Hit me up after your vacations. Enjoy! :smile:

swapagarwal:
Hey @sasek333, will you be willing to complete this? :smile:

shanmukhateja:
Actually @swapagarwal something's been bothering me for a while.  The only use of ping module is to check if the internet connection is working or not. 

And this condition is already satisfied when we are using messenger right? because it's only then we can use messenger right?

Am I missing something or is there something else to it?

swapagarwal:
@shanmukhateja Well, this would check whether a website can be reached or not, say is github.com down or not?

shanmukhateja:
Oh OK so it was meant for another website's working state. Good :)

VarunWachaspati:
@swapagarwal :If the intention is just to check whether a website is up or down, then we can request for the page from our host server if it returns a response then it is up right else if it times out for say continuously three packets then it must be down.

This way I guess cross compatibility will be maintained and purpose will also be served. If any other use case comes to your mind for which the above does not cover let me know.

swapagarwal:
@VarunWachaspati Yes, we can implement this on our server itself, but that adds to the maintenance cost!

Wert1996:
May I take this up? How about implementing this using isitup.org to check whether a website is down or not?

swapagarwal:
Added in #197

PR #33, opened 2016-05-06
added imdb link

AChelikani:
Added link for IMDb using 'ID' parameter of API response.


swapagarwal:
Thanks @AChelikani! :+1: 
As you have multiple PRs merged now, I suggest you move on to larger issues (we could really use some help there) and leave low hanging fruits for newbies. :smile:

PR #35, opened 2016-05-09
improve jokes module

shanmukhateja:
This commit improves the joke module by commenting out the API calls and using only a set of jokes.

- The jokes are present at **modules/src/jokes.txt**.
- config.py has a variable **JOKES_SOURCE_FILE** to access the file path.


swapagarwal:
Can you convert the text file to json format? That'd be much easier to parse (no splitting stuff required).

modules/src/joke.py
```diff
@@ -1,14 +1,30 @@
...
         data = r.json()
         output['input'] = input
         output['output'] = TextTemplate(data['joke']).get_message()
-        output['success'] = True
+        output['success'] = True        
+        '''
+        
+        with open(config.JOKES_SOURCE_FILE,"r") as file:
+            dump = file.read() # read the whole file
+            jokes_list = dump.split("\'''")[2].split("\n")
```
swapagarwal: This is quite hacky! We can make this easy by storing jokes in json format. :smile:

modules/src/joke.py
```diff
@@ -1,14 +1,30 @@
...
+            output['input'] = input
+            joke = choice(jokes_list).strip() #randomly choose a joke
+            if(len(joke)>2): #because there could be null values in the list
```
swapagarwal: why null values? it shouldn't be the case

modules/src/joke.py
```diff
@@ -1,14 +1,30 @@
...
+                output['output'] = TextTemplate(joke).get_message()
+            else:
+                output['output'] = TextTemplate(choice(jokes_list).strip()).get_message() #choose again
```
swapagarwal: what if you get the same joke again? (whose len <= 2)

modules/src/jokes.txt
```diff
@@ -0,0 +1,24 @@
+'''
+This file contains of all jokes to be served by JARVIS
+
+Format:
+
+<question>\t\t\t<answer>\n
+             (OR)
+<entire_joke_here>\n
+
+The "\t\t\t" splits the joke as question & answer( if its Q&A type) and \n is used for new line.
```
swapagarwal: I didn't get how and where splitting is being done. Can you please elaborate?

modules/src/joke.py
shanmukhateja: If there is a joke not following the above syntax, it might lead to a '' when split happens.

modules/src/jokes.txt
```diff
@@ -0,0 +1,24 @@
...
+
+'''
+
+What happens to a frog's car when it breaks down?\t\t\tIt gets toad away. 
+Why was six scared of seven?\t\t\tBecause seven "ate" nine.
+How do astronomers organize a party?\t\t\tThey planet.
+Want to hear a Potassium joke?\t\t\tK.
+A photon walks into a hotel. The desk clerk says, "Welcome to our hotel. Can we help you with your luggage?"\t\t\tThe photon says, "No thanks, I'm traveling light."
+What did the 30 degree angle say to the 90 degree angle?\t\t\t"You think you're always right!"
```
swapagarwal: nice selection! :+1:

modules/src/joke.py
swapagarwal: but you're adding the jokes in that format only, right?

modules/src/joke.py
shanmukhateja: I can assure you this may seem hacky but its actually not. It's become like this because of the comment block in the jokes.txt file. I've essentially put a comment block for newcomers.

modules/src/joke.py
shanmukhateja: Yes I put it keeping in mind of newcomers when the project horizons start expanding. Its just a \n character that when given extra can lead to a '' string.

modules/src/joke.py
shanmukhateja: Actually its not likely gonna appear as that's hoe the choice model makes choices. 
Explaining further, the null values only appear when there is multiple \n character. (Different text editors issues)
The probability is quite little for a null value to appear. I've tested it myself before PR

modules/src/joke.py
swapagarwal: ok, consider doing something like this:
rename jokes.txt to jokes.json and add the following data in it:
```
{
	"description": "the comment block for beginners",
	"jokes": [
		"joke #1",
		"joke #2",
		"... and so on..."
	]
}
```

modules/src/joke.py
swapagarwal: json will take care of that

modules/src/jokes.txt
shanmukhateja: 1. Every joke is split by a \n
2. Inside the joke string, it can be any model like a classic Q/A model or a one-liner.
3. In the final bot, question and answer will be clubbed as one with a \t\t\t where [0]th element of this split will be question and [1] would be the answer.

modules/src/joke.py
swapagarwal: still probability > 0
no problem, json will take care of that!

modules/src/jokes.txt
swapagarwal: is the 3rd point already implemented or will that be done in a future PR?

modules/src/jokes.txt
shanmukhateja: Thank you :)

modules/src/jokes.txt
shanmukhateja: The 3rd point needs to be implemented depending upon how the output is shown. I do not know how would the data be shown in the final output hence I've not tweaked it.

modules/src/joke.py
shanmukhateja: So apart from this, is there anything else @swapagarwal

swapagarwal: i think this should handle most of the issues.

modules/src/jokes.txt
swapagarwal: np

shanmukhateja:
@swapagarwal is there anything else to be done? May I know why hasn't this merged to master yet?
Am I missing something?

swapagarwal:
@shanmukhateja I missed this notification somehow!

shanmukhateja:
Well then why don't you bookmark it and check it manually? (I've for forgetfulness and even I do that)

swapagarwal:
Thanks @shanmukhateja for the PR! :+1: 
(Well, Github provides notifications but as I've subscribed to many of them, this one got lost.)

Issue #26, opened 2016-04-27
Replace links with buttons


swapagarwal:
- [x] book
- [x] movie
- [x] request
- [x] wiki

FleetAdmiral:
Hey could i take up this issue?

swapagarwal:
@FleetAdmiral Sure, go ahead!

UshalNaidoo:
#111 PR for button link on a book

nubianMONK:
@swapagarwal seems like all Low-Hanging-Fruit issues that are fantastic for a newbie as my self, have all been scooped up.  Are you planning on creating a newer list of the aforementioned?

SabareeshIyer:
Hey, is this open?

swapagarwal:
@nubianMONK Thanks for taking an interest! Yes, I'll be adding a few more low-hanging fruits. You might want to keep an eye out for them.
@SabareeshIyer Sorry, this is closed now. You can work on other issues / new additions.

PR #39, opened 2016-05-14
Migrate wiki (text -> button) (Issue #26)

conormag94:
JSON output for the Wiki module now follows the button template, similar to the movie and request modules.


swapagarwal:
Thanks @conormag94! :+1:

PR #50, opened 2016-05-23
Improve joke

badmon:
move jokes.json to data/* , modules should only contain major functionality of bot


swapagarwal:
Thanks @badmon! :+1: 
I was thinking along these lines too. :smile:

PR #58, opened 2016-06-01
Update anime.py

krizhan1997:
Format rating into two decimal places. 4.3233223 = 4.32
However, if the rating is 4.00, it becomes 4 rather than staying at 4.00 since 0 is redundant. But I can add it if u want.

### DIFF
--- a/modules/src/anime.py
+++ b/modules/src/anime.py
@@ -12,7 +12,7 @@ def process(input, entities):
 
         template = TextTemplate()
         template.set_text('Title: ' + data[0]['title'] + '\nSynopsis: ' + data[0]['synopsis'])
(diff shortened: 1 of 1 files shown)


swapagarwal:
A better option would be to use [`round`](https://docs.python.org/2/library/functions.html#round) :smile:

krizhan1997:
lol thanks thoo. hey i was wondering if there was a module i can try building for jarvis? just to get some hands on experience. anything u advice? i plan on working on it thoroughly since I finally have more free time and make sure it works just so I can better myself and get more experience. thanks for all the help tho

krizhan1997:
actually i will try working on the time module. if thats alright??

swapagarwal:
Sure, the low-hanging fruits are just meant to get started with the project. Now, you can move on to tackle larger issues. You can go ahead with any new module you like. If it's a one that's not mentioned in the README, it'd be better if you open an issue regarding the same. Hit me up if you need any help. Cheers!

swapagarwal:
Thanks @krizhan1997! :+1:

PR #59, opened 2016-06-04
Check Time Anywhere

krizhan1997:
eg.time in beijing or time in los angeles.
My first module. I feel like I may have messed up since its my first time and any feedback (even negative) regarding my code or even anything is much appreciated since its a learning experience.

### DIFF
--- a/modules/src/time.py
+++ b/modules/src/time.py
@@ -5,7 +5,7 @@ from datetime import datetime
 def process(input, entities):
     output = {}
     try:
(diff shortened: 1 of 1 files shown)


krizhan1997:
Just want to make sure this is alright.

swapagarwal:
Hey @krizhan1997, nice work! :smile: 
Seeing the [usage policy](http://wiki.openstreetmap.org/wiki/Nominatim#Usage_Policy), I think it's better to use MapQuest API.

krizhan1997:
ohh. ok np

swapagarwal:
Continued in #62

PR #62, opened 2016-06-05
New Api

krizhan1997:
You mean like this??


swapagarwal:
Yep, looks good!
Just add some tests and move the keys to config.py and this should be ready to merge.

swapagarwal:
Continued in #64

PR #65, opened 2016-06-06
Added error message for currency conversions

camdroid:


swapagarwal:
Hey @camdroid, it'd be better if you set the error message before the status.

camdroid:
Okay, that's fine. Out of curiosity, why?

swapagarwal:
Well, it should be consistent with the previous code (see dictionary module). Also, the status flag should ideally be set at the end! Hope that makes sense.

swapagarwal:
Thanks @camdroid! :+1:

PR #66, opened 2016-06-06
Added error message for finding anime.

camdroid:


swapagarwal:
Thanks @camdroid! :+1:

Issue #54, opened 2016-05-26
Add offline data sources


swapagarwal:
Below is the list of modules that can make use of an offline data source:

- fact
- joke
- quote

Add (Improve) the curated lists in `data/*` folder. Comment below if you'd like to take up one of these.

jorgeavaldez:
I can take this

swapagarwal:
sure @jorgeavaldez, let me know if you need any help! :smile:

nikhilgeo:
I like to take it up...

swapagarwal:
@nikhilgeo Sure, go ahead!

williamshoops96:
I've added some facts, about to pr now!

PR #73, opened 2016-06-07
added offline data for facts

williamshoops96:
as from #54


swapagarwal:
Thanks @williamshoops96! :+1:

PR #79, opened 2016-06-08
Weather Module

williamshoops96:
Sorry again about new pr, same reason - my last fork was a mess. No DS_store this time and fixed array error.


config.py
```diff
@@ -6,3 +6,4 @@
 JOKES_SOURCE_FILE = 'data/jokes.json'
 MAPQUEST_CONSUMER_KEY = '<MAPQUEST_CONSUMER_KEY>'
 TIME_ZONE_DB_API_KEY = '<TIME_ZONE_DB_API_KEY>'
+OPEN_WEATHER_MAP_ACCESS_TOKEN = '7af6de37fc4d3517129290b740663142'
```
andela-ijubril: I don't think having your access token in a config file is a good idea.

williamshoops96: It's a trash account anyways, just proof that it works:)

williamshoops96:
@swapagarwal Just been added to the testing list and the bot is better than I first expected. Super cool work man, and I'm so glad you've made it available to people like me!

I've just ran some local tests with my last commit and seems to be working good:)

swapagarwal:
Glad you like it!

swapagarwal:
Merged in 7fd41a12fd821bde70453b385061836785206ab2

PR #89, opened 2016-06-10
changed weather format

huzaifafaruqui:


swapagarwal:
Thanks @huzaifafaruqui! :+1:

PR #92, opened 2016-06-13
cleanup receipt

badmon:
remove if conditionals from code as its already handled by the use of defaults in functions ? what do you think ?


swapagarwal:
Hey @badmon, the `if` conditions are put in place so that the corresponding keys are set only when there is some value passed. Also, it denotes that they are optional. (Rest all keys are required!)

PR #97, opened 2016-06-20
Added basic bye functionality

rohshar:
Just so that Jarvis can respond if someone says bye


swapagarwal:
Thanks @rohshar! :+1:

Issue #31, opened 2016-05-01
Restaurant rating


havanagrawal:
Given the name of a restaurant/cafe/pub, find out locations and a (user) rating.

Nice to have:
1. Timings
2. Average bill for 2
3. Suggest location nearest to my current location
4. Specify cuisine

Ambitious:
1. Suggest 5 nearest restaurants based on my preference (eg: Veg, pizza, burger, italian, etc)

Sample queries:

1. How is the food at Starbucks?
2. Does Starbucks have good food?
3. Is Starbucks a good place to eat?
4. I'm hungry. Where's the nearest Starbucks?
5. I'm craving for some good food. How's Starbucks?
6. Do you know if Starbucks has good coffee?
7. Does Starbucks have cheap food?
8. What is the food rating for Starbucks.
9. I want to eat at Starbucks. Will it be open?

swapagarwal:
Truth be told, that'd be quite a useful module!
Do you have any APIs in mind where we could get this data from?
Also, the Messenger API doesn't provide us with location data (happy to be wrong here!) so we'd have to take it from the user.

havanagrawal:
This might be useful: https://developers.zomato.com/api

swapagarwal:
Looks good, any takers?

KartikKannapur:
I can help out. 
I have worked with their API earlier, but I didn't find it very stable .. Hopefully things are different now :)
I'll look into it.

swapagarwal:
Cool, that's great! :smile:

hithesh:
@KartikKannapur this looks really useful,  any improvements?? i would like to help

KartikKannapur:
@hithesh Sorry, I haven't had time to work on it. It would be great if you could take this up.

sudomichael:
I will have this completed within 2-3 days

katipallya:
So was this ever worked on furhter? @swapagarwal

swapagarwal:
@katipallya Sure, go ahead! You can create a test account for yourself to implement this; I'll create one for jarvis and link to config.py once the PR looks good.

ab9-er:
@swapagarwal What is this issue's status?

orbit-stabilizer:
Also would like to know

msultani:
I'm gonna go ahead and pick this up.

Nayan-Das:
I am interested.Can I take this if it is not solved yet.

msultani:
Yeah it’s all yours

PR #98, opened 2016-06-23
[wip] Add Restaurant Rating / Restaurant Suggestions, Fix Issue #31

108anup:
Could you give an idea of how to display the results for this query, I have written code to obtain the data from the ZOMATO API, I need to just return the data in appropriate format.

Also please train the wit.ai so that Testing could be done.

Pending Work:
Add tests in modules/tests
Return Formatted Output.


modules/src/__init__.py
```diff
@@ -16,4 +16,5 @@
     'video',
     'weather',
     'wiki',
+    'restaurant'
```
swapagarwal: alphabetical order + missing comma at end

swapagarwal:
@108anup Write the tests (they'll fail initially)
Return the output in Generic Template (see music module)

PR #100, opened 2016-07-03
Created quick reply template

MatthiasRMS:
Added class QuickReplyTemplate


swapagarwal:
Closing this as Quick Replies is not a template; rather it's an add-on to existing templates.
Added in 5128bf1a64463e7420c6b6cd29b5415f2240d921

PR #101, opened 2016-07-03
Created file attachment template

MatthiasRMS:


swapagarwal:
Thanks @MatthiasRMS! :+1:

Issue #104, opened 2016-07-11
Currency: Remove duplicate output


swapagarwal:
Query: `convert one usd to inr`
The output `1 USD = 67.1430 INR` is repeated two times!

PR #105, opened 2016-07-14
removed duplicate currency output, #104

bhague1281:
Changed `currency_conversion` string to only be set once, avoiding the duplicate currency output #104


swapagarwal:
@bhague1281 For the query `convert 39 usd to inr`, we still want to keep both the current outputs.

bhague1281:
@swapagarwal Changed the logic so that if the user enters "1", there's no duplicate.

swapagarwal:
Thanks @bhague1281! 👍

PR #106, opened 2016-08-05
Added coinflip feature

seongju:
The files are src/coinflip.py and tests/test_coinflip.py


modules/tests/test_coinflip.py
```diff
@@ -0,0 +1,7 @@
+import modules
+
+def test_coinflip():
+        assert('coinflip' == modules.process_query('Flip a coin')[0])
+        assert('coinflip' == modules.process_query('Jarvis flip a coin')[0])
+        assert('coinflip' == modules.process_query('Can you flip a coin')[0])
+        assert('coinflip' == modules.process_query('something random')[0])
```
swapagarwal: !=

modules/src/coinflip.py
```diff
@@ -0,0 +1,18 @@
+import random
+from templates.text import TextTemplate
+
+def process(input, entities):
+    output = {}
+    try:
```
swapagarwal: no need of try/catch as there is no api call, external file loading, etc. that can fail

modules/src/coinflip.py
```diff
@@ -0,0 +1,18 @@
...
+            output['input'] = input
+            if random.randrange(0,2,1):
+                    output['output'] = TextTemplate("Sure, it's heads!").get_message()
+            else:
+                    output['output'] = TextTemplate("Sure, it's tails!").get_message()
+
```
swapagarwal: remove extra blank line

swapagarwal:
I've left some comments, other than that it looks good 😄

swapagarwal:
Also, can you rename this module to coin?

swapagarwal:
Thanks @seongju! 👍

PR #107, opened 2016-08-06
Added Math functionality

JPHamlett1993:


swapagarwal:
Hey @JPHamlett1993, nice work! 😄 
Just a quick question, is the external library required?
IIRC, the `eval` function of python does something similar, can you check that out?

JPHamlett1993:
@swapagarwal I would say, yes. The module allows results for cos(45) and more math features than simply eval

swapagarwal:
Hmm, agreed but we should start with only the basic functions. Otherwise, we can go ahead and use the WolframAlpha API itself which can replace all our modules! 😜

JPHamlett1993:
@swapagarwal Ok, I have removed the module. It should do the basics now.

JPHamlett1993:
@swapagarwal What else needs to be done in order to merge this?

PR #117, opened 2016-08-24
pick random feature implemented

bhairavmehta95:
user writes "pick" + a bunch of options, and jarvis just picks one of the options for the user and returns it.

### DIFF
--- a/modules/src/__init__.py
+++ b/modules/src/__init__.py
@@ -17,4 +17,5 @@ __all__ = [
     'video',
     'weather',
     'wiki',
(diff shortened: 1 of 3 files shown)


swapagarwal:
As I understand, this is a generalized version of `coin` and `dice` modules, i.e. returning a random value out of a list of options. Can we merge them somehow or keep them independent?

bhairavmehta95:
Yeah, I will look into trying to merge this with the dice/coin functionality!

PR #118, opened 2016-09-06
Added New Feature - Sender Action

Sunil02324:
Added new feature - Sender Action to send mark seen and send typing actions to sender.
[Reference](https://developers.facebook.com/docs/messenger-platform/send-api-reference/sender-actions)


swapagarwal:
Jarvis replies pretty much instantly so I don't see the need for this.
Can you list some use cases where this might be useful?

PR #121, opened 2016-10-10
Quotes from api

netanel246:
Instead of reading quotes from a file, I get it from this [site](http://api.forismatic.com/api/1.0/) 
The benefits are simple.


netanel246:
Hi, There is something that I need to do in order to merge this?

swapagarwal:
Hey @netanel246, we actually moved the other way round, i.e. from using APIs to offline data sources. See #54 
Pros include fewer API calls, curated content, and filter out nsfw / offensive content.
If the API you are using has good content, we can filter it and add it offline. This is a one-time effort. Feel free to open another PR if you'd be willing to do this. 😄

netanel246:
Hey @swapagarwal , do you mean to create big file with the offline data?..
Maybe I can create generic script that creates offline data from online data...
What do you say about it?..

swapagarwal:
I'm not concerned with the size of the offline data; it can be small or big. It goes like this:
A contributor (like you) finds good content that we can add. Now, you can get that data from the API in any way you want, maybe a script. But after that data is put into a file, it needs to be reviewed manually before picking up the good ones and creating a PR.

PR #123, opened 2016-10-11
Add xkcd module to display most recent xkcd comic

conormag94:
XKCD comics are quite popular among programmers so I added a small module which displays the latest XKCD to the user. The api is found [here](https://xkcd.com/json.html) and is provided by the creator of the comic himself.

**Example queries:**
`Show me the latest xkcd`
`current xkcd`
`show me an XKCD`

Obviously I was unable to test it properly because there is no xkcd intent in the AI. However the module is small enough that I believe it should be working. I think I have followed all the Contributing guidelines but please let me know if I missed something.

### DIFF
--- a/modules/src/__init__.py
+++ b/modules/src/__init__.py
@@ -17,4 +17,5 @@ __all__ = [
     'video',
     'weather',
     'wiki',
+    'xkcd',
 ]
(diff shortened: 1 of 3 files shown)


conormag94:
@swapagarwal Could you please add some training in wit.ai for this?

swapagarwal:
@conormag94 Awesome idea for a module! 😄

swapagarwal:
Thanks @conormag94! 👍

PR #127, opened 2016-11-29
Improved Currency Module

AustinRow1:
I have implemented functionality that allows the user to use unofficial names for currencies. For example, rather than having to say 'USD to GBP' to get the conversion from US Dollars to British Pounds, the user can now say something like "usd to pounds" and get the same result. 

The names that can be used to refer to different currency symbols can be found in **data/currencies.json** which has been added to store this data. 

Note that until the AI has been trained more, there are some limitations to the usefulness of this. For example, the proper result will be returned if the user enters 'how much is us dollars to pounds', however, the AI will not be able to decipher the intent properly if given the request 'us dollars to pounds' or even more simply, 'us dollars to gbp'. 

The issues with the AI are not unique to this new functionality, however, as some requests that should already work, like 'usd to cad rate', will not work as the AI will not provide a to_currency in the entities that it returns. Though a similar request of 'usd to inr rate' will work. This just demonstrates a general need for more training in recognizing currency requests for the AI.


swapagarwal:
@AustinRow1 This looks like a clean, and quite useful addition.
However, as you correctly pointed out, the AI needs to be trained to support this.

data/currencies.json
```diff
@@ -0,0 +1,34 @@
+{
+  "USD": "USD", "DOLLAR": "USD", "BUCK": "USD", "US DOLLAR": "USD",
```
swapagarwal: Can there be a more concise way of storing this information?
I don't like USD being repeated.

AustinRow1: I wrote it this way so that it would always require constant time to check for a translation. I could implement it in the following format:
`{
    "USD": ["DOLLAR", "BUCK", "US DOLLAR"]
}`

This would make it more concise, but it would also slow it down. The slowdown would be negligible in times with little usage, but if Jarvis ever came under heavy load it may be nice to have faster translation.

What do you think?

swapagarwal: A quick search gave this: http://stackoverflow.com/questions/2974022
Can you explore more on this? The ideal solution would have a constant time lookup for a key, while removing the duplication of values.

swapagarwal:
Hey @AustinRow1, will you be willing to complete this? 😄

PR #133, opened 2016-12-08
Improve xkcd random posts

scottyla19:
Added random posts for xkcd. instead of newest post.

### DIFF
--- a/modules/src/xkcd.py
+++ b/modules/src/xkcd.py
@@ -7,27 +7,17 @@ from templates.text import TextTemplate
 def process(input, entities=None):
     output = {}
     try:
-        start_date = date(2016, 12, 7)
-        today = date.today()
-        start_num = 1769
-
-        mwf_count = 0
-        for d in range(1, (today - start_date).days):
-            cur_date = start_date + timedelta(days=d)
-            if cur_date.weekday() in [0, 2, 4]:
-                mwf_count = mwf_count + 1
-
-        max_num = start_num + mwf_count
+        r = requests.get('http://xkcd.com/info.0.json')
+        data = r.json()
+        max_num = data['num']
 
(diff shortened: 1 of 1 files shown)


swapagarwal:
Hey @scottyla19, this looks like a hacky way.
An easier way would be to get the latest comic, get its id, and then request for a random comic between 1 and latest. However, that would require 2 API calls.
Can you work on getting a cleaner solution, like given the date, find the latest id in the fastest way?

modules/src/xkcd.py
```diff
@@ -1,18 +1,34 @@
...
 from templates.generic import *
 from templates.text import TextTemplate
 
 def process(input, entities=None):
     output = {}
     try:
-        r = requests.get('http://xkcd.com/info.0.json')
+        start_date = date(2016, 12, 7)
+        today = date.today()
+        start_num = 1769
```
swapagarwal: I don't like starting from a magic number. Better to start from 0, I guess.

modules/src/xkcd.py
```diff
@@ -1,18 +1,34 @@
...
+            cur_date = start_date + timedelta(days=d)
+            if cur_date.weekday() in [0, 2, 4]:
+                mwf_count = mwf_count + 1
+
+        max_num = start_num + mwf_count
+
+        r = requests.get('http://xkcd.com/%d/info.0.json' % randint(1, max_num) )
         data = r.json()
 
+
```
swapagarwal: Remove extra lines.

swapagarwal:
Can you try making two API calls multiple times? If the latency is small, we can go ahead with that implementation for now, since it's cleaner. The first API call would be cached anyway; someone is already working on that in another PR. 😄

swapagarwal:
Thanks @scottyla19! 👍

PR #134, opened 2016-12-08
added motivation.json

sarthak5:


data/motivation.json
```diff
@@ -0,0 +1,19 @@
+{
+	"motivation": [
```
swapagarwal: motivations

swapagarwal:
Rename file to motivations.json

swapagarwal:
Have you gone through all of quotes, or do you need help? Also, remove them from quotes.json

sarthak5:
Yeah I've gone through all quotes and found these motivational. I'll remove these from quotes.json and change name to motivations.json.

swapagarwal:
Thanks @sarthak5! 👍

PR #137, opened 2016-12-14
random password generator utility

devanshdalal:


devanshdalal:
@swapagarwal, how does this work? How do I test my module.

devanshdalal:
Hi guys, anyone please respond. I deserve a hint at the least.

swapagarwal:
Hi @devanshdalal, can you provide a use case for this module?

devanshdalal:
Thanks @swapagarwal, AI assistant seems much cooler if it can reply to instant answers like this one. I don't know exactly whats your mission for JARVIS. Check a similar instant answer on [duckduckgo](https://duckduckgo.com/?q=random+password&t=hs&ia=answer). I don't know, maybe someone might wanna a completely random passwords like these for your FB,Quora etc, since their own passwords are crackable. I didn't think much about the use case, just created the pull request back then. But I am willing to contribute on this project in future. Hope it helps.

Also tell me how can I test my module? It is showing ``All checks have failed``. Do I have to add keywords for the module in my WIT account or something like this.

devanshdalal:
@swapagarwal , was the use case I gave, not good enough?

swapagarwal:
@devanshdalal Sorry for the late reply. 😇 
I totally agree that an AI assistant should be able to answer such queries.
But, I don't see people using a messenger bot for passwords. There are security issues, and a separate desktop / mobile app / browser plugin is best for this use-case.
The mission, should you choose to accept it, is to get more people involved into open-source, and keep it simple. That's it! 😄 
I'm willing to incorporate this in future if we see people requesting this.

horrormyth:
This seems to have gone stale @swapagarwal @devanshdalal

swapagarwal:
@edadesd @horrormyth What do you think?

edadesd:
I am very wary of sending users something and saying it's useful as a password. Serious security issues there. If it gets intercepted or logged, the password is compromised. Also potential legal issues since JARVIS will have handled the password.

Maybe the basic function of generating a random string could be adapted. A lottery ticket can be thought of as a string of random numbers, while a Scrabble tile set can be thought of as a string of random letters. Or maybe something like turning the user's query into a cryptogram for entertainment purposes.

horrormyth:
Sorry for the late intrusion here @swapagarwal  as @edadesd mentioned, it would be better not to have explicit password generator utility, in any case, that's a complete NO NO !!. I agree with the entertainment purpose, but a the same time, I am thinking of its value. Is it really needed with having all these ambiguities? I would rather not have it.

Issue #128, opened 2016-12-04
Add images in coin and dice module


swapagarwal:
As pointed out in #114,
Instead of displaying text, we can show an image or a gif of a coin showing the result (heads/tails).
This can be used in the dice roll as well.
See the [attachment](https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/templates/attachment.py) template. Use the Image type.

mahim97:
I was trying this out and could get this kind of output, 
{"text": "heads", "quick_replies": [{"payload": "{\"entities\": null, \"intent\": \"coin\"}", "attachment": {"attachment": {"type": "Image", "payload": {"url": "http://coins.silvercoinstoday.com/wp-content/uploads/2010/10/America-the-Beautiful-Silver-Coin-Obverse.jpg"}}}, "content_type": "text", "title": "Flip again!"}]}

Is this fine, I am finding it difficult to test it on messenger, need some help with figuring out ACCESS_KEY.

swapagarwal:
@mahim97, this is being worked on in #141

PR #141, opened 2016-12-18
Add images in coin and dice module #128

tdraebing:
Hi,
I changed the output of the coin module to give back an attachment with an image of a US quarter. The images were taken from wikimedia commons and are under public domain rights, so using them should be fine as I understand.
For the dice module I had the idea to use the unicode characters representing the dice sides, which should be more efficient. I wasn't able to test whether Facebook messenger would convert the unicode escape into the right symbol, but according to this quora thread (https://www.quora.com/Can-Messenger-Bots-reply-with-emoji-How) it should work. If there is a way to test it iin facebook messenger, I am happy to do it, but as I am new to using the facebook api I would need a hint. Thanks.

### DIFF
--- a/modules/src/coin.py
+++ b/modules/src/coin.py
@@ -1,10 +1,14 @@
 import json
 import random
-from templates.text import TextTemplate
+from templates.attachment import AttachmentTemplate
 from templates.quick_replies import add_quick_reply
 
+#images by US Mint; published on wikimedia under public domain rights.
+coin_images = {'heads': 'https://upload.wikimedia.org/wikipedia/commons/thumb/a/a0/2006_Quarter_Proof.png/244px-2006_Quarter_Proof.png',
+               'tails': 'https://upload.wikimedia.org/wikipedia/commons/thumb/7/70/Oregon_quarter%2C_reverse_side%2C_2005.jpg/236px-Oregon_quarter%2C_reverse_side%2C_2005.jpg'}
+
 def process(input, entities=None):
-    toss = TextTemplate(random.choice(['heads', 'tails'])).get_message()
+    toss = AttachmentTemplate(random.choice(list(coin_images.values())), type='image').get_message()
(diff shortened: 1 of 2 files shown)


swapagarwal:
Hey @tdraebing, nice work man!
I tested it and both modules work. I was thinking, maybe we can add an animation to the coin toss? What do you suggest?

tdraebing:
Hi @swapagarwal,
happy to hear that it works :-). Having an animation would be awesome. I had a look around the web, but could not actually find a good and freely usable gif or similar. Shutterstock has some good ones, but they are expensive. And honesty I just lack the skills in graphics stuff to do my own. I'll definitely will keep my eyes open, whether I can find one and would definitely do the implementation.
I made some notes of how I would imagine the animation to look like:

- If possible, there shouldn't be a visible background (or at least only a monochrome one). (that looks kind of cheap and wouldn't fit JARVIS personality)
- The animation shouldn't be too long. That might be annoying for the user (<10s).
- A gif might not be suitable, since it would repeat itself.
- We would need two animations (heads / tails) and they should be matching from the design.

swapagarwal:
@tdraebing Bull's eye!
So, we need matching animations with no background preferably under 5 seconds (even 10 might be too much I think!).
I'm not sure whether we can stop animations from looping.
We can go ahead with merging this for now, and creating a new issue for animations; or keep this open if you'd like to explore more. Your call?

tdraebing:
I think merging it for now makes sense. It is a step to the final aim. 
Getting the animations might be a lengthier process. I would for sure look further into it. In a new issue thread I could at least already implement the code needed to insert animations, to explore some possibilities like using gifs or other formats, so that adding animations would be a matter of changing the urls/paths as soon as we have a usable set.

swapagarwal:
Thanks @tdraebing! 👍

Issue #132, opened 2016-12-04
Book: Add author name


swapagarwal:

swapagarwal:
@mahim97 This was already taken up by @jmcomber. Feel free to take up any of the existing issues or a new addition. 😄

PR #135, opened 2016-12-09
Add author to book query

jmcomber:
Hi, I added the author in book queries. I tried for a few, and seems to work properly. 
It's a good thing that Anonymous/Unknown is a proper author according to GoodReads :).
Hope it's good. Didn't add anything to tests/test_book.py because I'm not sure how that works.

PR #142, opened 2016-12-20
Book: Add author name #132

mahim97:
I cant understand why a duplicate file of book.py was created, it don't have this file locally. Other than that everything seems to working fine.


swapagarwal:
The extra file is a temporary backup of book.py
If you observe the name correctly, it has a ~ at the end, denoting a copy. This is created by editors like vim when you open a file for editing.

swapagarwal:
Merged in #135

PR #147, opened 2016-12-31
Add new feature for latest news

harshitdx29:
News bot for handling requests of type:
-latest news
-news
-world news


swapagarwal:
@harshitdx29 The overall structure looks good! We would need to add attribution also, right?
Regarding the changes you made to GenericTemplate, are you sure `description` is supported? I couldn't find it here: https://developers.facebook.com/docs/messenger-platform/send-api-reference/generic-template

modules/src/news.py
```diff
@@ -0,0 +1,34 @@
+import requests
+import config
+import os
+import json
```
swapagarwal: required?

modules/src/news.py
```diff
@@ -0,0 +1,34 @@
...
+from templates.generic import *
+from templates.text import TextTemplate
+
+NEWS_API_KEY = os.environ.get('NEWS_API_KEY', config.NEWS_API_KEY)
+
+def process(input, entities):
```
swapagarwal: Since we are not using `entities`, we can initialize it to `None` to signify that.

modules/src/news.py
```diff
@@ -0,0 +1,34 @@
...
+    output = {}
+    source = 'google-news'
+    try:
+        r = requests.get('https://newsapi.org/v1/articles?source=' +  source + '&apiKey=' + NEWS_API_KEY)
+        data = r.json()
+        assert(len(data["articles"])>0)
```
swapagarwal: spacing around >

modules/src/news.py
```diff
@@ -0,0 +1,34 @@
...
+        template = GenericTemplate()
+        for article in data['articles']:
+            title = article['title']
+            description = article['description']
+            url = article['url']
+            template.add_element(title= title, item_url= url, description= description)
```
swapagarwal: remove extra space after =

modules/src/news.py
```diff
@@ -0,0 +1,34 @@
...
+        output['input'] = input
+        output['output'] = template.get_message()
+        output['success'] = True
+    except:
+        error_message = 'I couldn\'t perform that action.'
+        error_message += '\nPlease ask me something else, like:'
+        error_message += '\n  - latest news'
+        error_message += '\n  - world news'
+        error_message += '\n - news'
```
swapagarwal: align - with above lines

templates/generic.py
```diff
@@ -8,6 +8,7 @@
 BUTTON_TITLE_CHARACTER_LIMIT = 20
 BUTTON_LIMIT = 3
 ELEMENTS_LIMIT = 10
+DESCRIPTION_CHARACTER_LIMIT = 160
```
swapagarwal: source?

templates/generic.py
```diff
@@ -26,9 +27,10 @@ class GenericTemplate:
     def __init__(self):
         self.template = copy(template['value'])
         self.elements = []
-    def add_element(self, title='', item_url='', image_url='', subtitle='', buttons=[]):
+    def add_element(self, title='', description='', item_url='', image_url='', subtitle='', buttons=[]):
         element = {}
         element['title'] = title[:TITLE_CHARACTER_LIMIT]
+        element['description'] = description[:DESCRIPTION_CHARACTER_LIMIT]
```
swapagarwal: Need to check if messenger supports this.

harshitdx29: My Understanding was that GenericTemplate is something we have written as a wrapper for formatting the output we are generating for each module. Can you please tell me about it?

Also, I din't understand the meaning of adding attribution.

Would be great if you could please help with the above doubts :)

swapagarwal: So, messenger has some templates that it supports. I've created classes mapping those so that you only have to create a object with the required data, and the rest is taken care of (formatting the data as you see while interacting with the bot). The Generic Template is here: https://developers.facebook.com/docs/messenger-platform/send-api-reference/generic-template
As we are using third-party APIs for data, most of them are free to use but require us to add something like 'Powered by NewsAPI'. This check ensures that it's not our data to show, and we're just serving the data. This helps them gain audience. See the book / weather modules for reference.

harshitdx29:
Review changes done:
Majorly
1) Attribution was added as a button
2) Description was replaced by subtitle

harshitdx29:
Good to go ? 😆

swapagarwal:
Thanks @harshitdx29! 👍

harshitdx29:
Thanks @swapagarwal  👍  Should i update the README for features in a separate PR?

swapagarwal:
I have some backlog in README updation which I plan to do on weekend. 😇

PR #151, opened 2017-01-03
Add command line support for JARVIS

Icyphox:
I thought adding a command line version of JARVIS would interest the *nix userbase. 

So the basic idea for this is you have a script that runs JARVIS in the background, and when ever you want to query it, you just open a terminal and enter your query. This is what I'm doing right now:
jarvisscreen

I still need to add some kind of shell script that you can have JARVIS in `/bin`, y'know so you can make calls to JARVIS without having to `cd` to the directory and run it using Python.

What are your thoughts?

### DIFF
--- a/jarvis_cmd.py
+++ b/jarvis_cmd.py
@@ -1,0 +1,33 @@
+import urllib
+import ast
+import signal
+import sys
+
+ascii_art = """\
+
+
+  jjj                        iii      
+        aa aa rr rr  vv   vv      sss 
+  jjj  aa aaa rrr  r  vv vv  iii s    
+  jjj aa  aaa rr       vvv   iii  sss 
+  jjj  aaa aa rr        v    iii     s
+jjjj                              sss 
+
+Press Ctrl+C to quit
+"""
+print(ascii_art)
+
+while True:
+	try:
+		user_input = raw_input('query> ')
+	
+		url = "http://localhost:5000/search/?q=" + user_input
+		# opening the url
+		output_raw = urllib.urlopen(url).read()
+		output_dict = ast.literal_eval(output_raw)
+		print(output_dict["text"])
+
+	except KeyboardInterrupt:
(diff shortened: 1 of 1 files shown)


lakshaykalbhor:
@Icyphox  Rather than having a shell script for JARVIS in ```/bin```, a setup.py file could be added with entry point to the function maybe?
```
entry_points={
        'console_scripts': ['JARVIS=jarvis.jarvis_cmd:main'],
      }
```

swapagarwal:
@Icyphox Looks cool! This is basically exposing jarvis as a service, instead of just a messenger bot.
Did you set up API keys for the modules that are not working? I can update README.md with instructions to do that.

Icyphox:
Sure please do. But, I don't think it's got to do with the API keys. Look at the code, I'm just using `urllib` to pull the text data from the webpage.

Icyphox:
@lakshaykalbhor I know what you're talking about, but I've never actually done it myself. I'll look into it.

swapagarwal:
@Icyphox Silly me! 😛 
Now I understand what you are talking about. I'm afraid you'll have to parse the output template-by-template. The text template is the easiest, but for others you need to extract the information from `output_dict` accordingly.

swapagarwal:
A command-line tool may take some time, but it's too good a use-case to pass! We need to make this happen. 🔥

Icyphox:
Sorry for the late reply, was busy with a lotta IRL stuff. So yeah, we should do this.

swapagarwal:
Yes, Heroku's free tier will do.
So there are templates in which jarvis returns response, see https://github.com/swapagarwal/JARVIS-on-Messenger/tree/master/templates
The data returned by jarvis (in one of the above templates) can be parsed as needed. See the template file to understand the response formats.

Icyphox:
Alrighty! I'll see what I can do.

Issue #153, opened 2017-01-04
Return (semi-)random gif using the giphy API


tdraebing:
Hi,

how about using the giphy API (https://github.com/Giphy/GiphyAPI) to get gifs? This could be fully random currently trending gifs or targeted searches. e.g. `JARVIS show me a gif of funny cats` would search the API with the search terms `[funny, cats]` and would return a random gif given back by the API response.

swapagarwal:
@tdraebing Can't say no to funny cat gifs! 😆

swapagarwal:
I've requested for a production key; till then you can test with the public beta key `dc6zaTOxFJmzC`. Let me know if you run into any issues! 😄

swapagarwal:
They replied saying that bots should use the public key only! So you can go ahead and set that in config.py

PR #155, opened 2017-01-06
Added a module that returns a random gif following search terms

tdraebing:
#153

@swapagarwal I added a new module that returns a random gif. If any search terms / tags are given, they will be used to filter the gifs. Thus you would have to train the wit.ai to find the intent 'gif' and optionally return an 'gif'-entity with the search terms as value. The script checks, if this entity exists and otherwise returns an unfiltered random gif.

There is one thing I am not happy with. The Terms of the Giphy API says we have to include their logo and a 'powered by Giphy' message. I currently solved this by using the generic template and adding another image element with the logo. This is probably pretty ugly. Especially since there has to be a title for every element, thus it is kind of cluttering the chat. Facebook itself solved in a much more elegant way: https://tctechcrunch2011.files.wordpress.com/2015/07/photo-04-07-2015-19-27-47.png?w=383&h=680 .
But I couldn't find a way to do that in the Messenger API docs. Do you have an idea?

Anyway this pull request would be enough to at least already train the wit.ai and see whether it works in general.


swapagarwal:
That's not a pretty hack. 😞 
Let's both look for better alternatives; how other messenger bots are doing this..? In Messenger API, I found this: https://developers.facebook.com/docs/messenger-platform/send-api-reference/quick-replies

tdraebing:
Most bots actually just ignore that term of use or just add the logo to their webpage.
I don't think that adding the image to the quickresponse will work, since it would be cropped and it wouldn't be pretty either. I was thinkiing about watermarking the gifs on the fly using the pillow lib of python: http://pillow.readthedocs.io/en/4.0.x/reference/Image.html?highlight=paste#PIL.Image.Image.paste 
Just don't know how expensive (in server ressources) that is, especially since we would have to do that for every frame of the gif.

Or we just add 'powered by Giphy' to the title and make the gif a link to the giphy homepage in addition to adding the logo to the README.md. They might be fine with this.

swapagarwal:
Watermarking would probably be expensive. I can add the required attribution in README.md, that looks like a cleaner option.

tdraebing:
Seems indeed to be the best solution for now. I have removed the logo from the module.
Thanks for your input :-)

PR #158, opened 2017-01-22
changed from search within title to search within lyrics

acquayefrank:


swapagarwal:
There are a lot of files changed. Ideally, it should be a one-liner change in one file only!

modules/__init__.py
```diff
@@ -3,7 +3,7 @@
 import os
 import requests
 import sys
-from src import *
+from .src import *
```
swapagarwal: Why `.`?

templates/button.py
```diff
@@ -1,6 +1,6 @@
 from copy import deepcopy as copy
 
-from text import TextTemplate
+from .text import TextTemplate
```
swapagarwal: same as above

templates/generic.py
```diff
@@ -1,6 +1,6 @@
 from copy import deepcopy as copy
 
-from button import ButtonTemplate
+from .button import ButtonTemplate
```
swapagarwal: same as above

.gitignore
```diff
@@ -60,3 +60,4 @@ target/
 
 #Ipython Notebook
 .ipynb_checkpoints
+*.sqlite
```
swapagarwal: remove this for now, to keep it clean

templates/generic.py
acquayefrank: It's mostly a python-3 thing http://stackoverflow.com/questions/12172791/changes-in-import-statement-python3

.gitignore
acquayefrank: The caching syetem is creating the .sqlite file. Or should I just ignore it locally.

templates/generic.py
swapagarwal: Jarvis runs on Python 2.7

.gitignore
swapagarwal: Ignore it locally for now, to keep the PR clean for others as reference.

swapagarwal:
Added in #192

PR #162, opened 2017-01-28
downgrade to html5lib version compatible with bs4.

dranknight09:
'module' object has no attribute '_base' error with html5lib version 1.04, so downgrade to 0.9999999


swapagarwal:
Would you know why others are not facing this issue?

PR #167, opened 2017-02-06
Add twitter module for JARVIS

ZuZuD:
A simple module for twitter.
You need a twitter [API](https://dev.twitter.com/rest/public) account and learn [wit.ai](https://wit.ai/ZuZuD/JARVIS-on-Messenger/) to understand "twitter".

Twitter return a tweet based on both popularity and relevance.
Picture displayed is the media of the tweet itself, or the icon of the Author.

You can try on heroku platform [here](https://jarviszz.herokuapp.com/search/?q=Twitter%20moto)
A picture can be seen [here](http://www.hostingpics.net/viewer.php?id=258009jarvistwitter.png)

Thank you Jarvis ;)


ZuZuD:
I modified my contribution with possibilities to switch, depending on the fetched request, between 2 API

"twitter topic " to look for popular tweet about a topic (search API)
"twitter @author" to look for the last tweet of a user (timeline API)

Also, the tests added are more accurate.

horrormyth:
Any update on this @ZuZuD @swapagarwal

ZuZuD:
Still waiting for reviews.
Could update if relevant to someone interests.

modules/src/twitter.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def getUrl(tweet):
+    """ getUrl by author or popular topic """
+
+    if tweet[0] == '@':
+        author = tweet[1:]
+        req = requests.get(url=TWI_USER_TIMELINE_API + author, headers={
+                           'content-type': 'application/json'}, params=params['timeline'], auth=get_oauth())
+        req = req.json()[0]
```
horrormyth: This might fail if response fails, let's check the response first :), it would be nice to name the variable response/resp so that it is relevant. :)

modules/src/twitter.py
```diff
@@ -0,0 +1,130 @@
...
+        tweet_id = req['id_str']
+        url = 'https://twitter.com/%s/status/%s' % (author, tweet_id)
+    else:
+        req = requests.get(url=TWI_URL_SEARCH_API + tweet, headers={
+                           'content-type': 'application/json'}, params=params['search'], auth=get_oauth())
+        req = req.json()['statuses'][0]
```
horrormyth: same as line 51

modules/src/twitter.py
```diff
@@ -0,0 +1,130 @@
...
+
+    return url, req
+
+
+def getTitle(req, tweet):
+    """ Build title with hashtags """
+
+    try:
+        hashtags = req['entities']['hashtags']
+        if len(hashtags) < 1:
```
horrormyth: how about we do
 ```python 
if not hashtags: 
´´´

modules/src/twitter.py
```diff
@@ -0,0 +1,130 @@
...
+            raise KeyError
+        title = ''
+        for tag in hashtags:
+            title += '#%s ' % tag['text']
+        title = title.strip()
+    except KeyError:
+        if tweet[0] == '@':
```
horrormyth: How about we use this, more readable :)
``` python
startswith()

**Comment by horrormyth** [comment]
Let's fix the conflicts first, please :)

File: modules/src/twitter.py (hunk 1 above)
Comment (ZuZuD) [reply modules/src/twitter.py]: This is catch by try/except lines 113 to 127.
Any non-managed exception on the tweet process are handle this way.
Could be relevant to try to catch earlier but only for log purposes, can't deliver an HTTP error processed by the server to the client side.

**Comment by ZuZuD** [comment]
Thanks for your feedback buddy @horrormyth !

File: modules/src/twitter.py (hunk 1 above)
Comment (horrormyth) [reply modules/src/twitter.py]: Hey @ZuZuD  thanks for pointing out, overlooked it, but yes. Quite not happy about how the exceptions are caught in the line you specified, that is very vague. However, all the modules are written like you have done, so no comments there. Good to go from my side, unless @swapagarwal  has something to add 👍

File: modules/src/twitter.py [hunk 5]
```diff
@@ -101,6 +101,10 @@ def getDesc(req):
     return desc
 
 
+def startwith(char):
```
horrormyth: Hey, I overlooked this again, not a blocker, that you have constructed your own method for it but I was talking about the python's built in one, let's use the built one, please :) 
Reference 

- https://docs.python.org/3/library/stdtypes.html
- http://python-reference.readthedocs.io/en/latest/docs/str/startswith.html?highlight=startswith

ZuZuD: Oh my bad I didn't know this built-in function ;) I amended with an updated version. All my apologizes!

PR #140, opened 2016-12-17
Quick replies for jokes module

CoderOwl:
Added quick replies for one more joke and a quote for the jokes module. Tests not added. Functionality can be tested at https://www.facebook.com/devRevanthJARVIS/

Issue #131, opened 2016-12-04
Add quick replies to joke, fact and quote modules


swapagarwal:
Refer https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/modules/src/coin.py to add a quick reply (show me another, one more, etc.) in:
- [x] joke
- [x] fact
- [x] quote

CoderOwl:
Pull request created at:
https://github.com/swapagarwal/JARVIS-on-Messenger/pull/140 for adding quick replies ("One More!" and "Show me a quote") in joke module.

srulih:
Hi I'm new to open source and i'd like to try this out. Is this project is still available and if so could you please provide some more details as to what I need to do.

swapagarwal:
@srulih You can use the other commits as reference to update the `quote` module.

chanondw:
pull request #211 for quick reply on quote

PR #179, opened 2017-03-20
Quick Replies for Facts Module

edadesd:
Based on #140 and #131 I added quick replies to the Facts module. Thanks to @CoderOwl for their demonstrative PR for the Jokes module!


swapagarwal:
Thanks @edadesd! 👍

PR #182, opened 2017-03-29
League Table feature

dgrah50:
This is my first time using GitHub or contributing to a community project!
I have written a module that can display the football league tables of the top 5 European leagues; England , Spain, Germany, Italy or France.  The module uses BeautifulSoup to scrape the data from Guardian Football.
I would like the wit.ai to feed in the name of the country using **entities** so that the module knows which league to retrieve from Guardian.
The module will currently fail checks as it has a dependency on the BeautifulSoup Library.
I would be thankful for any feedback.

### DIFF
--- a/modules/src/__init__.py
+++ b/modules/src/__init__.py
@@ -11,6 +11,7 @@ __all__ = [
     'hello',
     'help',
     'joke',
+    'leaguetable',
     'lyrics',
     'movie',
     'music',
--- a/modules/src/leaguetable.py
+++ b/modules/src/leaguetable.py
@@ -54,9 +54,9 @@ def process(input, entities):
                
             
         holding+= ("Powered by Guardian Football")
-            output['input'] = input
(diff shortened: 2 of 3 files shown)


rennhak:
> This is my first time using GitHub or contributing to a community project!

Thank you for contributing!

conormag94:
@swapagarwal @dgrah50 If you are still looking for an API for this, I have used this API for a project in the past and found it very good. http://www.football-data.org/index

It has a league table API call as well which could easily be used.

swapagarwal:
@dgrah50 Would you be willing to complete this?

dgrah50:
Yes I can do this

horrormyth:
@dgrah50 Looks like this has gone stale :)

addyvm22:
Hello i have written the code with an api call for this feature...Please check it out over here... https://github.com/dgrah50/JARVIS-on-Messenger/pull/1
@dgrah50 @swapagarwal

PR #185, opened 2017-05-21
Fetch weather from user's current location

jamesloyys:
Hi there!

This is my first ever pull request and I hope that I'll be able to contribute something of value.

This pull request fetches weather from the user's current location - allows TRAVIS to provide weather information in the context of the user's location, without requiring the user to explicitly state a location.

Example: "What's the weather now?", "What's the weather like today?"

### DIFF
--- a/modules/src/weather.py
+++ b/modules/src/weather.py
@@ -7,23 +7,34 @@ MAPQUEST_CONSUMER_KEY = os.environ.get('MAPQUEST_CONSUMER_KEY', config.MAPQUEST_
 OPEN_WEATHER_MAP_ACCESS_TOKEN = os.environ.get('OPEN_WEATHER_MAP_ACCESS_TOKEN', config.OPEN_WEATHER_MAP_ACCESS_TOKEN)
 
 def process(input, entities):
-    output = {}
-    try:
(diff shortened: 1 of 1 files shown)


modules/src/weather.py
```diff
@@ -7,23 +7,34 @@
...
+	output = {}
+	try:
+		if entities != {}:
+			r = requests.get('http://open.mapquestapi.com/nominatim/v1/search.php?key=' + MAPQUEST_CONSUMER_KEY + '&format=json&q='+ entities['weather_location'][0]['value'] + '&limit=1')
+			location_data = r.json()
+			r = requests.get('http://api.openweathermap.org/data/2.5/weather?lat='+ location_data[0]['lat'] + '&lon='+ location_data[0]['lon'] + '&units=metric&appid=' + OPEN_WEATHER_MAP_ACCESS_TOKEN)
+			weather_data = r.json()
+			location_display_name = location_data[0]['display_name']
+		else:
+			r = requests.get('http://ipinfo.io/json')
```
swapagarwal: Wouldn't this retrieve the IP of the server where jarvis is hosted?
A better way would be to prompt user for his location. 
See https://developers.facebook.com/docs/messenger-platform/send-api-reference/quick-replies and https://developers.facebook.com/docs/messenger-platform/webhook-reference/message

PR #190, opened 2017-05-22
Extract "youtubeVideoId" from the API response, and add another link to anime.py

guzmanga:


modules/src/anime.py
```diff
@@ -17,7 +17,8 @@ def process(input, entities):
 
         template = TextTemplate()
         template.set_text('Title: ' + top_anime['canonicalTitle'] + '\nSynopsis: ' + top_anime['synopsis'])
-        template.set_post_text('\nAverage Rating: ' + top_anime['averageRating'] + '%')
+        template.set_post_text('\nAverage Rating: ' + top_anime['averageRating']
+                                + '\nYoutube Video: https://www.youtube.com/watch?v=' + top_anime[youtubeVideoId] + '%')
```
swapagarwal: Link should be added a button. See https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/modules/src/music.py#L27

swapagarwal:
Closed in #194

Issue #188, opened 2017-05-21
Anime: Add YouTube link


swapagarwal:
Extract "youtubeVideoId" from the [API response](https://kitsu.io/api/edge/anime?filter[text]=naruto), and add another link [here](https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/modules/src/anime.py#L24).

ActualSize:
Could I do this one, I've never contributed before and this seems like a good starting point?

guzmanga:
I would also like to try this one, if it's ok.

sanjitjain2:
Hello, I'm interested, I am a first timer and would take some time to understand the code. Is this issue still open?
I would like to help.
Can someone guide me?

shivam043:
Sorry sir,forgot to write the comment..
I have added a youtube link, Can you verify it??

PR #194, opened 2017-06-01
Anime: Add YouTube link #188

shivam043:
#188 Added youtube link for the anime.py file

### DIFF
--- a/modules/src/anime.py
+++ b/modules/src/anime.py
@@ -22,6 +22,7 @@ def process(input, entities):
 
         template = ButtonTemplate(text)
         template.add_web_url('Kitsu URL', 'https://kitsu.io/anime/' + top_anime['slug'])
+        template.add_web_url('Youtube URL', 'https://youtube.com/' + top_anime['youtubeVideoId'])
 
         output['input'] = input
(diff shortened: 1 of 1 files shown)


swapagarwal:
Hey @shivam043, it would be great if you could comment on the issue before starting to code so that work is not duplicated. 😅  See #190

modules/src/anime.py
```diff
@@ -21,7 +21,7 @@ def process(input, entities):
         text = template.get_text()
 
         template = ButtonTemplate(text)
-        template.add_web_url('Kitsu URL', 'https://kitsu.io/anime/' + top_anime['slug'])
+        template.add_web_url('Kitsu URL', 'https://kitsu.io/anime/' + top_anime['slug'],'Youtube URL', 'https://www.youtube.com/watch?v=' + top_anime['youtubeVideoId'])
```
swapagarwal: There should be two buttons. See https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/modules/src/music.py#L27

swapagarwal:
Thanks @shivam043! 👍

PR #197, opened 2017-06-09
Feature to check whether a website is running or not

Wert1996:
Checks whether a web domain is up or not using isitup.org. Needs to learn 'Is xyz.com up', 'xyz.com status'.


modules/src/__init__.py
```diff
@@ -21,6 +21,7 @@
     'url',
     'video',
     'weather',
+    'webcheck',
```
swapagarwal: Can we rename it to ping module? It's more intuitive.

modules/src/webcheck.py
```diff
@@ -0,0 +1,29 @@
...
+from templates.text import TextTemplate
+
+def process(input, entitites):
+    output = {}
+    try:
+        url = entities['url'][0]['value']
+        strip_list = ['http://','https://','/']
+        for item in strip_list:
+            url = url.strip(item)
+        page = requests.get('https://isitup.org/'+str(url))
```
swapagarwal: Pro Tip: See what happens if you append .json to the end of the URL..

Wert1996:
Thanks for the feedback! Applied the proposed changes :)

modules/src/ping.py
```diff
@@ -0,0 +1,35 @@
...
+def process(input, entitites):
+    output = {}
+    try:
+        url = entities['url'][0]['value']
+        orig_url = url
+        strip_list = ['http://','https://','/']
+        for item in strip_list:
+            url = url.strip(item)
+        page = requests.get('https://isitup.org/'+url+'.json')
+        page = json.loads(page.text)
```
swapagarwal: If you see the other modules, the general structure is:
r = requests.get(...)
data = r.json()

modules/src/ping.py
```diff
@@ -0,0 +1,35 @@
...
+    except:
+        error_message = 'There seems to be some problem'
+        error_message = '\nPlease ask me something else, like:'
+        error_message = '\n - is google.com up'
+        error_message = '\n - google.com status'
+        error_message = '\n - ping google.com'
+        output['error_msg'] = TextTemplate(error_message).get_message()
+        output['success'] = False
+    return output
+
```
swapagarwal: Remove extra line.

modules/src/ping.py
```diff
@@ -0,0 +1,35 @@
...
+        elif status == 2:
+            txt = orig_url + ' seems to be down!'
+        else :
+            txt = 'I need a valid domain to check!'
+        output['input'] = input
+        output['output'] = TextTemplate(txt).get_message()
+        output['success'] = True
+    except:
+        error_message = 'There seems to be some problem'
+        error_message = '\nPlease ask me something else, like:'
```
swapagarwal: It should be "+=". Check the below lines too.

modules/src/ping.py
```diff
@@ -0,0 +1,35 @@
...
+        url = entities['url'][0]['value']
+        orig_url = url
+        strip_list = ['http://','https://','/']
+        for item in strip_list:
+            url = url.strip(item)
+        page = requests.get('https://isitup.org/'+url+'.json')
+        page = json.loads(page.text)
+        status = page['status_code']
+        if status == 1:
+            txt = orig_url + ' is up!'
```
swapagarwal: Don't use original url here, as that might be misleading. Use the one that we passed in the api call.

modules/src/ping.py
```diff
@@ -0,0 +1,35 @@
...
+        elif status == 2:
+            txt = orig_url + ' seems to be down!'
+        else :
```
swapagarwal: Do we know the status code here? If yes, then we should use that here instead. Any unexpected status codes should result in an error.

modules/src/ping.py
```diff
@@ -0,0 +1,35 @@
...
+
+def process(input, entitites):
+    output = {}
+    try:
+        url = entities['url'][0]['value']
+        orig_url = url
+        strip_list = ['http://','https://','/']
+        for item in strip_list:
+            url = url.strip(item)
+        page = requests.get('https://isitup.org/'+url+'.json')
```
swapagarwal: spacing around +

modules/src/ping.py
```diff
@@ -0,0 +1,35 @@
...
+import json
+
+from templates.text import TextTemplate
+
+def process(input, entitites):
+    output = {}
+    try:
+        url = entities['url'][0]['value']
+        orig_url = url
+        strip_list = ['http://','https://','/']
```
swapagarwal: Add a space after comma. Can you explain why the third entry in the list is required?

Wert1996: Coz queries like google.com/ will cause wrong domain name error.

Wert1996:
Thanks @swapagarwal  :)

swapagarwal:
Congratulations! 🎉 
Good work on your first pull request. 😄

PR #201, opened 2017-06-15
Minor code cleaning, formatted error messages

williamsonz:


williamsonz:
I wasn't sure on the best approach. For now just cleaning it up for easier future updates.
I'll hit it again tonight and see what I can do. Sorry bout that ^^,

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

modules/src/error_msg.py
```diff
@@ -0,0 +1,55 @@
+
```
swapagarwal: This isn't a module. It's more of a helper class. Can be moved to modules/examples.py
Keep the base error in the module itself for now, and extract just the list of examples out. This should be used both for the error messages in src files, as well as in tests.

modules/src/help.py
```diff
@@ -1,35 +1,36 @@
...
-    help += '\n  - videos of sia'
-    help += '\n  - flip a coin'
-    help += '\n  - roll a die'
-    help += '\n  - show a random xkcd comic'
-    help += '\n  - latest news'
-    help += '\n  - paradise lyrics'
-    help += '\n\nI\'m always learning, so do come back and say hi from time to time!'
-    help += '\nHave a nice day. :)'
+            helper = helper.replace('there', entities['sender']['first_name'])
+    helper += """
```
swapagarwal: Reset this file and others which are related to triple-quoting the strings. They can be picked up in a separate PR.

swapagarwal:
Also, rebase your work on the current master to remove conflicts.

modules/src/help.py
williamsonz: Covered in the initial pull-request; I will include it with the next one,

modules/src/error_msg.py
williamsonz: Unsure, so asking for clarification:
Split into two files, one for error messages (error_msg.py) and one for examples (examples.py)?
I understand storing them a directory above, to use for both tests and src files.

swapagarwal: Let's make only one file now: examples.py. (error_msg.py can be created later if required; let the error messages be in the corresponding module itself, like it is now. Will visit it later)
examples file would have a list of sample queries for each of the modules, which will be used in both src and tests.

modules/src/help.py
swapagarwal: Remove these changes for now. Only keep the examples refactoring in this PR.

horrormyth:
@williamsonz Could we fix these, please.

Issue #203, opened 2017-06-17
Ping: Handle URLs with path / parameters


swapagarwal:
Some queries that are currently failing:
- ping google.com/maps
- ping google.com?a=b

Only the hostname should be checked for.

PR #204, opened 2017-06-18
Fix: Handle URLs with path / parameters (#203)

LucasBPM:
Added url parsing to extract hostname from url

### DIFF
--- a/modules/src/ping.py
+++ b/modules/src/ping.py
@@ -1,14 +1,15 @@
 import requests
 
 from templates.text import TextTemplate
+from urlparse       import urlparse
 
 def process(input, entities):
     output = {}
     try:
         url = entities['url'][0]['value']
-        strip_list = ['http://', 'https://', '/']
-        for item in strip_list:
-            url = url.strip(item)
+        if '//' not in url:
+          url = '%s%s'%('//', url)
+        url = urlparse(url).hostname
         r = requests.get('https://isitup.org/' + url + '.json')
         data = r.json()
         status = data['status_code']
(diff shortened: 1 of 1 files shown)


modules/src/ping.py
```diff
@@ -1,14 +1,15 @@
...
 def process(input, entities):
     output = {}
     try:
         url = entities['url'][0]['value']
-        strip_list = ['http://', 'https://', '/']
-        for item in strip_list:
-            url = url.strip(item)
+        if '//' not in url:
+          url = '%s%s'%('//', url)
+        url = urlparse(url).hostname
```
swapagarwal: `hostname` returns `None` if it is unable to parse the url. See https://docs.python.org/2/library/urlparse.html Throw an error in that case.

modules/src/ping.py
```diff
@@ -1,14 +1,15 @@
...
+from urlparse       import urlparse
 
 def process(input, entities):
     output = {}
     try:
         url = entities['url'][0]['value']
-        strip_list = ['http://', 'https://', '/']
-        for item in strip_list:
-            url = url.strip(item)
+        if '//' not in url:
```
swapagarwal: This does not look elegant. Can be made better by trying to get the scheme of the parsed url, and adding it if does not exist; then parse it again.

modules/src/ping.py
```diff
@@ -1,14 +1,15 @@
...
+          url = '%s%s'%('//', url)
```
swapagarwal: Use 4 spaces for indentation.

LucasBPM:
Not too sure if I handled the exception properly but the rest of it should be good.

swapagarwal:
Thanks @LucasBPM! 👍

PR #213, opened 2017-06-25
Create runtime.txt for easy heroku usage.

hafiz703:
Based on https://devcenter.heroku.com/articles/python-runtimes. By default it generated with python 3 hence creating problems when trying to deploy. This file would hopefully avert that


swapagarwal:
Looks like a recent change: https://devcenter.heroku.com/changelog-items/1178

swapagarwal:
Tested by deploying to Heroku, with and w/o the runtime.txt file.
Thanks @hafiz703! 👍

PR #218, opened 2017-06-27
Upcoming programming contests added

torquecoder:
Some of the sample queries are: 

- upcoming cp contests
- programming contests
- cp contest
- competitive programming contests

### DIFF
--- a/modules/src/__init__.py
+++ b/modules/src/__init__.py
@@ -4,6 +4,7 @@ __all__ = [
     'book',
     'bye',
     'coin',
+    'cpcontest',
     'currency',
     'dice',
     'dictionary',
--- a/modules/src/cpcontest.py
+++ b/modules/src/cpcontest.py
@@ -1,0 +1,31 @@
+import requests
+import os
+from templates.text import TextTemplate
+
+
+def process(input, entities):
+    output = {}
+    try:
+
+        r = requests.get('https://cpcontest-api.herokuapp.com/')
+        data = r.json()
+
+        template = GenericTemplate()
+        buttons = ButtonTemplate()
+
+        # Number of results to be displayed
+        num = 3
+        for contest in data['result']['upcoming_contests']:
+            buttons.add_web_url(contest['Name'], contest['url'])
+            num -= 1
+            if num == 0:
+                break
+
+        output['input'] = input
+        output['output'] = buttons.get_message()
+        output['success'] = True
+    except:
+        error_message = 'There was some error while retrieving data.'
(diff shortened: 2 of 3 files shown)


torquecoder:
Any update?

modules/src/__init__.py
```diff
@@ -4,6 +4,7 @@
     'book',
     'bye',
     'coin',
+    'cpcontest',
```
swapagarwal: Let's think of a better name.

modules/src/cpcontest.py
```diff
@@ -0,0 +1,31 @@
+import requests
+import os
+from templates.text import TextTemplate
+
+
+def process(input, entities):
+    output = {}
+    try:
+
+        r = requests.get('https://cpcontest-api.herokuapp.com/')
```
swapagarwal: Is this your personal api? If yes, do you plan to maintain it (hosting, bug fixes, etc.)?

modules/src/cpcontest.py
```diff
@@ -0,0 +1,31 @@
...
+        template = GenericTemplate()
+        buttons = ButtonTemplate()
+
+        # Number of results to be displayed
+        num = 3
+        for contest in data['result']['upcoming_contests']:
+            buttons.add_web_url(contest['Name'], contest['url'])
+            num -= 1
+            if num == 0:
+                break
```
swapagarwal: A better way to do this would be to specify this number in the API call itself.

modules/src/cpcontest.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def process(input, entities):
+    output = {}
+    try:
+
+        r = requests.get('https://cpcontest-api.herokuapp.com/')
+        data = r.json()
+
+        template = GenericTemplate()
```
swapagarwal: Unused?

modules/src/cpcontest.py
torquecoder: Yes, this is my personal API and I am maintaining it. If I could find any alternative API, which is very well maintained and serves the purpose, I will update it.

modules/src/__init__.py
torquecoder: "contest" would be too generalized, other names are too big such as "programingcontest" or something like that. So, I abbreviated "competitive programming" to "cp" here. Do you have any suggestion?

horrormyth:
@torquecoder  Let's fix the tests please, one of the tests is  failing, :)

PR #223, opened 2017-06-29
Use image with square aspect ratio in music

### LINKED ISSUE #209: Images with square aspect ratio
Previously, `GenericTemplate` only rendered images in the ratio 1.91:1. Now, we have the option of rendering it in 1:1 ratio! Check all modules where `GenericTemplate` is being used, and see jarvis' response to determine whether we should switch to square images in that module.
e.g. `ping` module uses square aspect ratio.

SubhrajyotiSen:
Resolves #209

### DIFF
--- a/modules/src/music.py
+++ b/modules/src/music.py
@@ -16,6 +16,7 @@ def process(input, entities):
             data = r.json()
         assert (len(data['tracks']['items']) > 0)
         template = GenericTemplate()
+        template.set_image_aspect_ratio_to_square()
         for track in data['tracks']['items']:
             title = track['name']
             item_url = track['external_urls']['spotify']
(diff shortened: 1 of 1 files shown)


swapagarwal:
Hey @SubhrajyotiSen, you mentioned that this PR resolves the related issue. Have you checked the aspect ratio in all relevant modules?

SubhrajyotiSen:
Hey, I have checked the modules where GenericTemplate is used and found that only in the music module, the images used have an aspect ratio of 1:1 since it's album art. Other modules like news and xkcd have different images aspect ratio

swapagarwal:
Thanks @SubhrajyotiSen! 👍

PR #231, opened 2017-07-04
Updated GET request in anime.py module to only pull one result

### LINKED ISSUE #208: Optimize network calls
Some of the API calls can be optimized by limiting the response.
e.g. If XYZ API returns 10 results by default, and we use only the top result; we can specify (in the request) how many items to return in the response, which optimizes the amount of data that gets transferred.
First Step: Figure out all such API calls.

NolanNicholson:
From what I could see, the anime.py module was the only one pulling more than one search result. All the other API calls that could return multiple results already had an "&limit=1" flag. This should resolve #208 .


swapagarwal:
Thanks @NolanNicholson! 👍

Issue #224, opened 2017-06-29
music module is down


swapagarwal:
See https://developer.spotify.com/migration-guide-for-unauthenticated-web-api-calls/
Also, we needed #207 resolved yesterday!

edadesd:
I think we'll need to use the Client Credentials flow, because we don't make requests on behalf of a particular user that require access to that user's account. The Client Credentials flow is explained here: https://developer.spotify.com/web-api/authorization-guide/#client_credentials_flow 

We'll need to add a line for the API key to config.py and rework the module to get an access token, then carry out the request authenticating with the token.

edadesd:
I'm working on this at the moment.

PR #232, opened 2017-07-04
Music auth

edadesd:
The music module was failing because all requests now need to be authenticated, as mentioned in #224 . I made the module authenticate via the Client Credentials flow, because we don't need to authorize on behalf of the user.

To accommodate this patch, @swapagarwal will need to register the app with Spotify at: https://developer.spotify.com/my-applications/

Then, generate the key for config.py by base64-encoding the client_id and client secret together in the following format, without the brackets:

[client_id]:[client secret]

I was able to get this to work using the encoder at https://www.base64encode.org/ . This PR also adds a line to config.py for the Spotify key.


edadesd:
Made the changes requested by Stickler.

edadesd:
Squashed commits to clean up history.

modules/src/music.py
```diff
@@ -10,9 +12,19 @@
...
+
+        r = requests.post(auth_url, headers=headers, data=payload)
+
+        key = r.json()['access_token']
+
         music = entities['music'][0]['value']
         with requests_cache.enabled('music_cache', backend='sqlite', expire_after=3600):
-            r = requests.get('https://api.spotify.com/v1/search?q=' + music + '&type=track')
+            headers = {'Authorization': 'Bearer ' + key}
+            r = requests.get('https://api.spotify.com/v1/search?q=' + music + '&type=track&limit=1', headers=headers)
```
swapagarwal: Why limit = 1?

modules/src/music.py
```diff
@@ -10,9 +12,19 @@
...
     output = {}
     try:
+
+        auth_url = "https://accounts.spotify.com/api/token"
+        headers = {'Authorization': 'Basic ' + config.SPOTIFY_API_KEY}
+        payload = {'grant_type': 'client_credentials'}
+
+        r = requests.post(auth_url, headers=headers, data=payload)
+
+        key = r.json()['access_token']
```
swapagarwal: Hmm, this token would be fetched everytime? Let's think of avoiding this somehow.

swapagarwal:
Wow, you really have provided detailed instructions. Thanks Daniel!

modules/src/music.py
edadesd: Based on #208 I assumed we only used the first result. But looking at the rest of the module, I guess not.

modules/src/music.py
edadesd: Maybe save the token in a file, attempt to re-use it. If that fails, repeat the query and request a new token.

edadesd:
Pushed a commit to remove the limit to one response.

modules/src/music.py
edadesd: Better yet, we can save it in config.py to avoid needing to create any new files.

edadesd:
I made it store the access token in a file (music_token.json) and attempt to use the stored access token at the start of each script run. If the API returns 401 for unauthorized, it asks for a new token and stores that in the file, then repeats the request with the stored token. This allows the module to use a retrieved token until it expires.

edadesd:
Rebased to current master.

edadesd:
Rebased to current master.

swapagarwal:
Thanks @edadesd! 👍 
I checked the module with the query "linkin park songs" and it's working as expected. 🎉

PR #234, opened 2017-07-05
Add additional bye greetings

FShiwani:
2 more greetings for the bye module.


horrormyth:
@swapagarwal This seems okay

swapagarwal:
Thanks @FShiwani! 👍

Issue #207, opened 2017-06-20
Status Dashboard


swapagarwal:

swapagarwal:
A `status` module can be put in place, which will show the corresponding statuses fetched from *somewhere* (updated every hour or so). Need to explore options for storing the information. Simple options include caching, cron job, etc.

edadesd:
Maybe a JSON file with a dictionary filled with keys = API name, values = status. Then a cron job that runs a script every hour to poll for status and update the JSON file if needed.

PR #235, opened 2017-07-10
Status script for #207

atalkar:
Base implementation of endpoint-status observation. Makes module tests accessible. Runs LL module tests as determinant of module health and outputs result to json file. As suggested in #207, status.py could be run as cron job, updating status.json every 5 minutes. 

This is a starting implementation. I'd like input from the core maintainers about what to do with this status information. Should there be special messages for downed endpoints? Should downed endpoints not be displayed in the help message?


swapagarwal:
Hey @atalkar, you're checking for the current tests. However, they are incomplete, and only test the classification of the query to the relevant module. The deep check should also include whether the backend serving the module is up or not.

atalkar:
So, would this involve an additional test for each module that hits each supported endpoint and makes sure we don't get a 5xx HTTP response? If so, would be be best to add a 'test_<module_name>_status' function to each module's test file and then have 'status.py' run all these tests?

PR #236, opened 2017-07-11
Tests for test_anime.py

edadesd:
1) Changed the README to note that the API for the anime module is named Kitsu, not Hummingbird.

2) Added some unit tests to test_anime.py to verify the items included in the anime module's response and their accuracy compared to another call to the base API.

There is a list of sample anime titles at the top of the file, which is shuffled for each run of the test suite. Most of the tests iterate across the list, comparing the module response for the title to the API response for the title at the relevant data members.


edadesd:
Closing this and setting up a new branch with the changes to squash the commits.

PR #238, opened 2017-07-12
More tests for anime module

edadesd:
Changed the README to note that the API for the anime module is named Kitsu, not Hummingbird.

Added some unit tests to test_anime.py to verify the items included in the anime module's response and their accuracy compared to another call to the base API.

There is a list of sample anime titles at the top of the file, which is shuffled for each run of the test suite. Most of the tests iterate across the list, comparing the module response for the title to the API response for the title at the relevant data members.


edadesd:
I messed up trying to update to current master. I'll close this and the other PR and remake one for both.

Issue #242, opened 2017-07-18
movie module is down


swapagarwal:
The backend API has gone private, and is no longer free.
https://www.patreon.com/posts/api-is-going-10743518
Look for an alternative.
P.S. It'd be great to have #207 resolved asap, rather than have people report that a module is down in the gitter chatroom! 😅

edadesd:
What do you think of: https://www.themoviedb.org/documentation/api ?

edadesd:
One disadvantage of using TMDb is that it doesn't generate an IMDB link the way OMDb did. I have a basic module patch for TMDb at: https://github.com/edadesd/JARVIS-on-Messenger/commits/movieAPI

a1shadows:
Why not use the official rotten tomatoes api? It's legal and just as legit as imdb.

edadesd:
Rotten Tomatoes and Fandango APIS: https://developer.fandango.com/

It sounds like the Rotten Tomatoes API only gives reviews and scores. The (related) Fandango API says it can give movie information, but from the documentation it doesn't appear that you can search by movie name. You have to search by location to get movies showing at nearby theaters.

a1shadows:
TMDb it is, then.

edadesd:
You can keep looking for another API if you want. I only linked to my TMDb patch instead of setting up a PR in case someone found a better API to use.

swapagarwal:
IMDb id is important, as we only provide a gist of the movie information in the response.
Here's a relevant link: https://www.themoviedb.org/talk/5851f2ddc3a368258c0127b0 😉

edadesd:
That's a good link, thanks for finding it. I'll work on a version that uses the TMDB ID to fetch the IMDB ID and makes an IMDB link.

edadesd:
Opened PR #251 to address this.

PR #251, opened 2017-07-31
Reworked movie module to rely on TMDB.

edadesd:
Addresses issue #242  by making the module rely on the TMDB API instead of OMDB. TMDB requires attribution by placing a logo and a non-endorsement message. The logo was placed in the movie module's source box in the README, while the non-attribution message is in the module's documentation.

@swapagarwal will need to sign up for a TMDB account for the app at: https://www.themoviedb.org/account/signup - the instructions for getting an API key can be found at https://www.themoviedb.org/faq/api?language=en


modules/src/movie.py
```diff
@@ -1,22 +1,46 @@
...
         movie = entities['movie'][0]['value']
+
+        # Make a search request to the API to get the movie's TMDB ID.
+        # TMDB required attribution notice: "This product uses the TMDb API but is not
+        # endorsed or certified by TMDb."
+
         with requests_cache.enabled('movie_cache', backend='sqlite', expire_after=86400):
-            r = requests.get('http://www.omdbapi.com/?t=' + movie + '&plot=full&r=json')
+            request_str = 'https://api.themoviedb.org/3/search/movie?api_key='
+            request_str += config.TMDB_API_KEY + '&query=' + movie
```
swapagarwal: It'd be cleaner to use `params` parameter in `requests.get()` method to generate the query string. You can refer to the `lyrics` module: https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/modules/src/lyrics.py#L17-L20

swapagarwal:
@edadesd I've signed up for a TMDb account, and added the api key on server. 😄

swapagarwal:
Also, 👍 for following the attribution guidelines!

edadesd:
Made the requested changes.

swapagarwal:
Thanks Daniel! 🎉

Issue #215, opened 2017-06-25
Gender-neutral responses


swapagarwal:
Some messages are gender-specific, like:
```
I'm sorry; I'm not sure I understand what you're trying to say sir.
Try typing "help" or "request"
```
Can (Should) we make them gender-neutral?

leegenes:
+1
Don't see added value in having a gendered answer, and do see potential for cases to detract from the bot's value for some users.

edadesd:
+1
It's a good thing to strive for.

ssummun54:
What about asking in the beginning for the sex and then he goes off of that. We could have a gender neutral option too

_Sent from my Huawei Nexus 6P using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

swapagarwal:
We can get the gender from the User Profile API. 🎉 
Just like we're getting the name [here](https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/modules/__init__.py#L48), we can get more info about the [user](https://developers.facebook.com/docs/messenger-platform/user-profile).

leegenes:
Safest bet seems to be using name, if that's available.

pt2pham:
The gender is available in the API, names could include names that are applicable to both genders. 

_Sent from my Samsung SM-G930W8 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

leegenes:
@pt2pham - I guess I'm suggesting we skip gender. Just refer to people by name.

pt2pham:
@leegenes ah I see, had a misunderstanding. That seems to be the best solution I'd agree.

_Sent from my Samsung SM-G930W8 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

nicholassavini:
I agree with avoiding the concept of gender altogether. Some users may be equally turned off it were only presented as a binary option, and it seems to add little value to include gendered answers.

sync350:
Is anyone actively working on this? Looks like the general consensus is to avoid using genders, and to refer to the user by name instead. I'd love to try my hand at this. It's my first time working with this kind of project, so I can't make any promises, but I'm willing to give it a shot.

swapagarwal:
Based on the above discussions, the consensus is to avoid the concept of gender altogether.

PR #253, opened 2017-08-01
Gender-neutral responses #215

sync350:
Edited the file to remove "sir" from the response if the bot is unable to
respond to the query. This still leaves room for someone to add a replace
command to get the sender_name variable, and allow the bot to address the
user by name (such as within the other modules), but for right now this seems
to be the best way to stay gender neutral.
I was looking at using a replace function, but so far have had no such luck.

### DIFF
--- a/modules/__init__.py
+++ b/modules/__init__.py
@@ -61,4 +61,4 @@ def search(input, sender=None, postback=False):
                 return TextTemplate('Something didn\'t work as expected! I\'ll report this to my master.').get_message()
     else:
         return TextTemplate(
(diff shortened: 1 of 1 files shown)


aviral1701:
@swapagarwal @sync350 I would like to do the additional work of fetching the name variable and displaying it out!

sync350:
@aviral1701 @swapagarwal  That sounds fine to me :)

aviral1701:
Let's just for @swapagarwal to respond on this one,then we can start the new work!

swapagarwal:
Thanks @sync350! 👍 
@aviral1701 You may go ahead. 🏃

PR #255, opened 2017-08-02
Added support for both new and old API's response structure

ahmedsadman:
The newest version of wit.ai uses a different response structure. In this new structure, "intent" is NOT a separate key, instead it is integrated in "entities". 

So, when new contributors teaches the WIT ai to work with their module, they will face problems because JARVIS uses older api version which uses different response structure.

### DIFF
--- a/modules/__init__.py
+++ b/modules/__init__.py
@@ -29,7 +29,6 @@ def process_query(input):
         entities = data['outcomes'][0]['entities']
         confidence = data['outcomes'][0]['confidence'] if 'intent' not in data['outcomes'][0] \
             ['entities'] else data['outcomes'][0]['entities']['intent'][0]['confidence']
-        print intent, confidence
         if intent in src.__all__ and confidence > 0.5:
             return intent, entities
         else:
--- a/modules/__init__.py
+++ b/modules/__init__.py
@@ -24,9 +24,10 @@ def process_query(input):
             'Authorization': 'Bearer %s' % WIT_AI_ACCESS_TOKEN
         })
         data = r.json()
-        intent = data['outcomes'][0]['intent']
+        intent = data['outcomes'][0]['intent'] if 'intent' not in data['outcomes'][0]['entities'] else data['outcomes'][0]['entities']['intent'][0]['value']
         entities = data['outcomes'][0]['entities']
-        confidence = data['outcomes'][0]['confidence']
+        confidence = data['outcomes'][0]['confidence'] if 'intent' not in data['outcomes'][0]['entities'] else data['outcomes'][0]['entities']['intent'][0]['confidence']
+        print intent, confidence
         if intent in src.__all__ and confidence > 0.5:
             return intent, entities
         else:
--- a/modules/__init__.py
+++ b/modules/__init__.py
@@ -24,9 +24,11 @@ def process_query(input):
             'Authorization': 'Bearer %s' % WIT_AI_ACCESS_TOKEN
         })
         data = r.json()
-        intent = data['outcomes'][0]['intent'] if 'intent' not in data['outcomes'][0]['entities'] else data['outcomes'][0]['entities']['intent'][0]['value']
+        intent = data['outcomes'][0]['intent'] if 'intent' not in data['outcomes'][0] \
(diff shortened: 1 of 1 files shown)


modules/__init__.py
```diff
@@ -24,9 +24,10 @@ def process_query(input):
             'Authorization': 'Bearer %s' % WIT_AI_ACCESS_TOKEN
         })
         data = r.json()
-        intent = data['outcomes'][0]['intent']
+        intent = data['outcomes'][0]['intent'] if 'intent' not in data['outcomes'][0]['entities'] else data['outcomes'][0]['entities']['intent'][0]['value']
         entities = data['outcomes'][0]['entities']
-        confidence = data['outcomes'][0]['confidence']
+        confidence = data['outcomes'][0]['confidence'] if 'intent' not in data['outcomes'][0]['entities'] else data['outcomes'][0]['entities']['intent'][0]['confidence']
+        print intent, confidence
```
horrormyth: Do we need this?

modules/__init__.py
```diff
@@ -24,9 +24,10 @@ def process_query(input):
             'Authorization': 'Bearer %s' % WIT_AI_ACCESS_TOKEN
         })
         data = r.json()
-        intent = data['outcomes'][0]['intent']
+        intent = data['outcomes'][0]['intent'] if 'intent' not in data['outcomes'][0]['entities'] else data['outcomes'][0]['entities']['intent'][0]['value']
```
horrormyth: Could we follow PEP8 please,  https://www.python.org/dev/peps/pep-0008/#id19

horrormyth: If we could follow Could we follow PEP8 please, https://www.python.org/dev/peps/pep-0008/#id19 then that would be greate

modules/__init__.py
horrormyth: Do we need this

swapagarwal:
@ahmedsadman The data-training is done on wit console (independent of the version), and the code that calls Wit API includes the version (in jarvis). I'm not sure I understand how new users will face problems. Can you elaborate?
@horrormyth Thank you for taking the time to review! 🎁 I had included style-check bot a month back, but disabled it temporarily as it seemed to dishearten new contributors.

modules/__init__.py
ahmedsadman: Sorry, the print statement was for debugging purpose, I removed it in the latest commit.

modules/__init__.py
ahmedsadman: Commit done

astroengisci:
@swapagarwal I had the same problem when I was writing my dice extension. If you go into wit.ai, create a new bot, and sub in its access key, even though the request includes the version, the format is different. Somehow, "intent" is now included in the "entities" block of the JSON by default, rather than being it's own block right under "outcomes". Just like @ahmedsadman, I had to modify my local code for testing purposes. 

So, your wit.ai returns:

    "outcomes":
    {
        "intent":
        {
            // which module should handle it
        }
        "entities":
        {
            // stuff for the module to handle
        }
    }

However, with a new app made for testing purposes, it returns:

    "outcomes":
    {
        "entities":
        {
            "intent":
            {
                // which module should handle it
            }
            "some_other_entity":
            {
                // stuff for the module to handle
            }
        }
    }

swapagarwal:
@astroengisci Oh, it seems like new apps are getting the new structure. I suspect that since jarvis is an old app, it continues to get the old structure from wit to maintain backwards compatibility. I'll try creating a new app today, and see how it goes. Also, there must be some blog post mentioning these changes. That official link should have complete information on all the api changes in the new version.

ahmedsadman:
@swapagarwal  Thanks for the understanding. Please let me know if there is any update on this so that I can continue working on the soccer module. 

_Note: It's all up to you whether you want to create a new app (which is surely gonna take good amount of effort), but actually it's not necessary, because the code I submitted in this request can handle both new and old structure. In this way, new module creators would not face problem while using the latest version._

swapagarwal:
Opened https://github.com/wit-ai/wit/issues/697

PR #262, opened 2017-08-06
Fix channel results breaking Video search

FosterSamuel:
The video module should not abandon requests if a result is a channel link.

## Example Problem
If you say to JARVIS "linkin park videos", the last result is a channel link. This result causes JARVIS to abandon the search and default to the "I couldn't find any videos" message.

## Problem and Solution
Not all results from the YouTube API have `videoId` inside of `id`.

```python
item_url = 'https://www.youtube.com/watch?v=' + item['id']['videoId']
```

When the result is a channel link, it instead contains `channelId` inside of `id`. This pull request covers that reality with:
```python
try:
      # Video result from search
      item_url = 'https://www.youtube.com/watch?v=' + item['id']['videoId']
      buttons.add_web_url('YouTube Link', 'https://www.youtube.com/watch?v=' + item['id']['videoId'])
except:
      # Channel result from search
      item_url = 'https://www.youtube.com/channel/' + item['id']['channelId']      
```

Both results need a "Channel Link", but only one of them needs a "YouTube Link" button.

## Semantic Argument
Maybe a channel result when searching for videos is not right. The answer to this is to ignore channel results. If @swapagarwal wants that, I will go in that direction. 👍 

Regardless of direction, I do not believe the video module should give up on the search because of one channel result.

### DIFF
--- a/modules/src/video.py
+++ b/modules/src/video.py
@@ -21,11 +21,16 @@ def process(input, entities):
         template = GenericTemplate()
         for item in data['items']:
             title = item['snippet']['title']
-            item_url = 'https://www.youtube.com/watch?v=' + item['id']['videoId']
-            image_url = item['snippet']['thumbnails']['high']['url']
             subtitle = item['snippet']['channelTitle']
+            image_url = item['snippet']['thumbnails']['high']['url']
             buttons = ButtonTemplate()
-            buttons.add_web_url('YouTube Link', 'https://www.youtube.com/watch?v=' + item['id']['videoId'])
+            try:
+                # Video result from search
+                item_url = 'https://www.youtube.com/watch?v=' + item['id']['videoId']
+                buttons.add_web_url('YouTube Link', 'https://www.youtube.com/watch?v=' + item['id']['videoId'])
+            except:
+                # Channel result from search
+                item_url = 'https://www.youtube.com/channel/' + item['id']['channelId']
             buttons.add_web_url('Channel Link', 'https://www.youtube.com/channel/' + item['snippet']['channelId'])
             template.add_element(title=title, item_url=item_url, image_url=image_url, subtitle=subtitle,
                                  buttons=buttons.get_buttons())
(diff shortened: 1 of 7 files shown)


swapagarwal:
@FosterSamuel This is a really detailed write-up. Thanks for taking the time to deep-dive into the issue! 😄

modules/src/video.py
```diff
@@ -21,11 +21,16 @@ def process(input, entities):
...
+            image_url = item['snippet']['thumbnails']['high']['url']
             buttons = ButtonTemplate()
-            buttons.add_web_url('YouTube Link', 'https://www.youtube.com/watch?v=' + item['id']['videoId'])
+            try:
+                # Video result from search
+                item_url = 'https://www.youtube.com/watch?v=' + item['id']['videoId']
+                buttons.add_web_url('YouTube Link', 'https://www.youtube.com/watch?v=' + item['id']['videoId'])
+            except:
+                # Channel result from search
+                item_url = 'https://www.youtube.com/channel/' + item['id']['channelId']
```
swapagarwal: This might also result in an error, in case the result is of type `playlist`. See https://developers.google.com/youtube/v3/docs/search

FosterSamuel:
Gladly! 👍

Thank you for noticing the chance of a playlist result. I'll work on adding that soon! 

_Sent from my Google Nexus 6P using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

FosterSamuel:
Gladly! 👍

Thank you for noticing the chance of a playlist result. I'll work on adding that soon!  

_Sent from my Google Nexus 6P using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

FosterSamuel:
Gladly! 👍

Thank you for noticing the chance of a playlist result. I'll work on adding that soon!

FosterSamuel:
# Changes 
1. Removed `type=video` from API query to allow Playlist results 
2. Changed "YouTube Link" --> "Video Link" for clarity
3. Added code to handle all three types of results

swapagarwal:
If `type` was set to `video` before, how come `channel` was returned in response? 🤔

FosterSamuel:
@swapagarwal Not sure. Seems to be a less strict part of the YouTube API. 

If you use the first commit (with `type=video` in the query), it exposes the results of a search that may have channel results in it. The type only excluded playlists, for some reason. 

I can record a video example of this if you'd like further proof of the statements.

PR #265, opened 2017-08-09
wiki module:added images in case of Disambiguation

aviral1701:


aviral1701:
#265 fixed

aviral1701:
the test for time module fails! But I didn't even change the time module.

aviral1701:
Any Updates?

modules/src/wiki.py
```diff
@@ -40,7 +39,7 @@ def process(input, entities):
                     }
                 }
                 buttons.add_postback('Wikipedia Summary', payload)
-                template.add_element(title=data.title, item_url=data.url, image_url=image_url,
+                template.add_element(title=data.title, item_url=data.url, image_url=data.images[0],
```
swapagarwal: The `images` list might be empty. In that case, let's default to the current image_url.

aviral1701:
@swapagarwal Please review sir

swapagarwal:
Hey @aviral1701, I deployed this change to test it; but the output isn't that good. What do you think?

aviral1701:
@swapagarwal Ya result is not so good.May be we should stick with the wiki image only.

swapagarwal:
Can we get the main wiki image using some other API? If not, let's drop this.

swapagarwal:
Some relevant links I found:
https://www.mediawiki.org/wiki/Extension%3aPageImages
https://en.wikipedia.org/w/api.php?action=query&prop=pageimages&format=json&piprop=original&titles=facebook

aviral1701:
@swapagarwal Hey, sorry for late response,was busy with my midterms.
So I was thinking once we get the page IDs,we can make a simple request to the wikipedia api(Second link you mentioned) to fetch the main image.I have tested it and it works fine.

Issue #264, opened 2017-08-09
Add an api key wrapper


swapagarwal:
This extraction https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/modules/src/dictionary.py#L9 would be abstracted out to a util class.

edadesd:
I've started working on this. As it stands, I instantiate the class and then call a class method passing the name of the key or the data file variable name. The method then needs to retrieve the key from the environment, or failing that, from config.

edadesd:
Another way to do it would be to take the key name as a constructor argument and have the constructor store the key as an object variable which could then be retrieved from the object. 

Do you have any thoughts on which you would prefer?

luxorv:
Is this still an issue? I would like to work on this.

edadesd:
PR #266 is currently opened to address it. You could comment on that, or even open up your own PR if you want to implement it in a different way.

luxorv:
@edadesd seems like #266 solves the issue, I don't think more action is needed but to fix the test not passing in the code. Thanks for the quick response :)

Abdullah-alfadhel:
Hello, I was wondering if you want an Arabic translator? so the project can reach a wider audience.

PR #266, opened 2017-08-11
Add GetConfig class to abstract fetching keys/data files

edadesd:
Addresses Issue #264 by creating a GetConfig class in util.py that will fetch a key or data file from config.py. 

Create the object in a module and pass it the name of the key variable. The class keeps a dictionary mapping names to variables in the config file. Then retrieve the fetched value by accessing the GetConfig object's "result" variable. If you need another line from config.py, re-create the GetConfig object passing it the name of the next variable you want.


conormag94:
It might be a good idea to wrap the keys in config.py inside a dictionary (called "api_keys" or something) instead of having the dictionary inside this class. 

Then you could just use the class to load the key from config.api_keys. Any new API keys would only need to be added once in config.py instead of in two places.

edadesd:
@conormag94 Sounds good. And then once all of the modules have been updated to rely on the wrapper class, we can make the dictionary the base data structure for config.py.

edadesd:
The failure occurred in the test_book module, which this PR doesn't change at all. I wonder if the Wit AI isn't properly handling the query.
```

=================================== FAILURES ===================================

__________________________________ test_book ___________________________________

    def test_book():

        assert ('book' == modules.process_query('book timeline')[0])

        assert ('book' == modules.process_query('have you read harry potter?')[0])

>       assert ('book' == modules.process_query('little women book rating')[0])

E       assert 'book' == None

modules/tests/test_book.py:7: AssertionError
```

edadesd:
I'd like to give the keys in the config.py dict better mnemonic names, such as "music" so they're easier to recall when using them in a module. Does anyone think this would be helpful?

For example,
Old:
`'SPOTIFY_API_KEY': SPOTIFY_API_KEY,`

New:
`'music': SPOTIFY_API_KEY,`

edadesd:
Would it be preferable to have the wrapper class inside config.py to avoid the need for another file?

conormag94:
@edadesd Yeah I think it could probably go inside config.py for now. Unless @swapagarwal has any particular preference for this class.

edadesd:
Pushed a commit moving the wrapper class inside config.py

elpaxoudis:
Hello, I am trying to get involved in an open source project. I am trying to learn how all this ecosystem works here, so please be kind :)

Specifically, I stumbled upon this particular PR #266 which seems quite old. It seems like it's still open. Is that the case? I mean why is it not closed after almost 2 years?  Thanks in advance!

edadesd:
The basic reason is that people with commit access don't always have time to commit patches. I don't think the PR has been merged yet so the issue is still technically open. If you want to work on it, please feel free to!

elpaxoudis:
Cool, thanks for the quick reply. If I do work on it, I should make a new PR right? Referencing of course the issue and this PR.

edadesd:
Yeah, that would be a good way of handling it.

swapagarwal:
@edadesd 
HI Daniel
Thanks for your kind words towards the maintainers. I'll try to clean up issues / PRs this weekend to get ready for Hacktoberfest! 😇
Cheers.

Issue #260, opened 2017-08-06
Movie: Correct Rating


swapagarwal:
`happyness movie` returns Rating = 7.7
This rating is from TMDb: https://www.themoviedb.org/movie/1402-the-pursuit-of-happyness
IMDb rating is 8.0 (http://www.imdb.com/title/tt0454921/)

Options:
* Rename field "Average Rating" to "TMDb Rating"
* Retrieve and show IMDb rating

ahmedsadman:
I can work on this. There is no IMDB API available, so I want to do this: get the IMDB id from TMDB API, then use the IMDB id to construct a full IMDB website link, and scrape the Rating. Is that okay?
**And yes, can I work on this module while the WIT AI response problem still persists?** I think the AI is taught in the older version style, right?

swapagarwal:
Let's explore more options (avoid scraping). The Wit response structure update doesn't affect jarvis, for old as well as new modules. The effect is seen when you create a **new app** for testing purposes.

ahmedsadman:
Well, do you have any other option in mind? Please let me know.
And if wit response doesn't affect new modules, how do you propose to use WIT if I want to continue working on the Soccer module? I need to create new app so that I can train it about user inputs, right? And Soccer module is new of course. Frankly speaking, I am highly confused at this stage: At one time, you're saying that new modules are not affected, again, you're saying that effect is seen only when creating new apps. **So, how contributors can create new modules without creating new apps?**

swapagarwal:
New modules (soccer) in old app (jarvis) are not affected.
If you create new app (say jarvis-test), it will have new response structure.
While submitting a Pull Request, you only submit the module code. I'll train the new module (soccer) in old app (jarvis), so it'll have the old response structure. Hope that makes sense! 😅

ahmedsadman:
Oh! That's what I was asking for days. When I asked for help, some people in Gitter told me that I need to create own app in WIT to make module. I never knew that you were going to train that for me. 
It was the only barrier between me and soccer module, I will start on that right away!

swapagarwal:
So you can create your own app for testing purposes, till your module is integrated. Sorry for the misunderstanding! 😞

ahmedsadman:
Okay, about the IMDB rating: you already know that OMDB provides api that can fetch IMDB rating, what if we use OMDB just for fetching the Rating?

swapagarwal:
[...]

PR #268, opened 2017-08-11
Added ability to fetch IMDB rating

ahmedsadman:
Based on #260,
Now the Movie module can show both IMDB and TMDB ratings.


ahmedsadman:
The check failed because I imported a new module named "imdb". How can I fix this?

modules/src/movie.py
```diff
@@ -33,9 +36,16 @@ def process(input, entities):
             })
             data = r.json()
 
+        # fetch the rating from IMDB
+        imdb_id = data['imdb_id']
+        imdb_search = ia.search_movie(imdb_id)
```
swapagarwal: The `search_movie` method is useful when we are querying the movie using keywords. Here, we already have the IMDb ID. We can directly use the `get_movie` method here.

modules/src/movie.py
```diff
@@ -33,9 +36,16 @@ def process(input, entities):
...
+        imdb_search = imdb_search[0]
+        ia.update(imdb_search)
+
         template = TextTemplate('Title: ' + data['title'] +
                                 '\nYear: ' + data['release_date'][:4] +
-                                '\nAverage Rating: ' + str(data['vote_average']) + ' / 10' +
+                                '\nTMDB Rating: ' + str(data['vote_average']) + ' / 10' +
```
swapagarwal: Let's keep only one rating (imdb) here.

modules/src/movie.py
ahmedsadman: Done. I didn't know about get_movie.

modules/src/movie.py
ahmedsadman: Done

swapagarwal:
Thanks @ahmedsadman! 👍

PR #269, opened 2017-08-11
Changed 'offline' sources in README to 'JARVIS'

edadesd:
I thought that referring to the sources with local data files as "Offline" unnecessarily suggested that the modules weren't operational. So I changed their sources to "JARVIS" to show that the outputs are coming from data files within this project.


swapagarwal:
Makes sense! 💯

Issue #259, opened 2017-08-06
Movie: Add trailer link


swapagarwal:
Refer the TMDb API documentation.

agiridharan:
Can I help out here? @swapagarwal

swapagarwal:
@agiridharan Sure, go ahead! 😄

jwetzell:
I believe I have done this if no progress has been made.

aviral1701:
@swapagarwal Can I make a PR to close this one quickly?

darthgera123:
is this issue still open?

PR #272, opened 2017-08-17
Added trailer link

agiridharan:
#259

### DIFF
--- a/modules/src/movie.py
+++ b/modules/src/movie.py
@@ -26,18 +26,24 @@ def process(input, entities):
             assert (len(data['results']) > 0)
             tmdb_id = str(data['results'][0]['id'])
             # Make another request to the API using the movie's TMDb ID to get the movie's IMDb ID
-            r = requests.get('https://api.themoviedb.org/3/movie/' + tmdb_id + '?api_key={api_key}&append_to_response=videos', params={
+            r = requests.get('https://api.themoviedb.org/3/movie/' + tmdb_id, params={
                 'api_key': TMDB_API_KEY
             })
             data = r.json()
+            videoRequests = requests.get('https://api.themoviedb.org/3/movie/' + tmdb_id + '/videos', params={
+                'api_key': TMDB_API_KEY
+            })
+            videoData = videoRequests.json()
+            with open('data.txt', 'w') as outfile:
+                json.dump(videoData, outfile)
         template = TextTemplate('Title: ' + data['title'] +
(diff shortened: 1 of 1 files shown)


modules/src/movie.py
```diff
@@ -23,20 +23,18 @@ def process(input, entities):
                 'include_adult': False
             })
             data = r.json()
-
             assert (len(data['results']) > 0)
             tmdb_id = str(data['results'][0]['id'])
-
             # Make another request to the API using the movie's TMDb ID to get the movie's IMDb ID
-            r = requests.get('https://api.themoviedb.org/3/movie/' + tmdb_id, params={
+            r = requests.get('https://api.themoviedb.org/3/movie/' + tmdb_id + '?api_key={api_key}&append_to_response=videos', params={
```
swapagarwal: These should be put in `params`.

modules/src/movie.py
```diff
@@ -23,20 +23,18 @@ def process(input, entities):
...
                 'api_key': TMDB_API_KEY
             })
             data = r.json()
-
         template = TextTemplate('Title: ' + data['title'] +
                                 '\nYear: ' + data['release_date'][:4] +
                                 '\nAverage Rating: ' + str(data['vote_average']) + ' / 10' +
-                                '\nOverview: ' + data['overview'])
+                                '\nOverview: ' + data['overview'] + 
+                                '\nTrailer: https://www.youtube.com/watch?v=' + data['videos']['results'][0]['key'])
```
swapagarwal: Add this as a button.

agiridharan:
Updated the movies.py file

modules/src/movie.py
```diff
@@ -23,23 +23,27 @@ def process(input, entities):
...
             assert (len(data['results']) > 0)
             tmdb_id = str(data['results'][0]['id'])
-
             # Make another request to the API using the movie's TMDb ID to get the movie's IMDb ID
             r = requests.get('https://api.themoviedb.org/3/movie/' + tmdb_id, params={
                 'api_key': TMDB_API_KEY
             })
             data = r.json()
-
+            videoRequests = requests.get('https://api.themoviedb.org/3/movie/' + tmdb_id + '/videos', params={
```
swapagarwal: Why did we move to making a separate call? Does `append_to_response` not solve our use-case?

swapagarwal:
Also, the merge conflict still exists. Can you rebase on current master branch?

swapagarwal:
Closing due to inactivity. Continued in #278

PR #276, opened 2017-08-20
Move spotify token to a folder

azie-ginanjar:
- add tokens folder
- move spotify_token.json to tokens folder
- change SPOTIFY_TOKEN_FILE path on config.py


swapagarwal:
Thanks @azie-ginanjar! 👍

PR #277, opened 2017-08-21
Added a recipe module

alexmotoc:
I tried testing it, but no query worked so far because the wit.ai is probably not configured to handle this new type. I added some sample queries in the test_recipe.py. I am looking forward to receiving feedback on my work and suggestions for improvement. :+1: 

Should I select a random recipe and display it or would something like the Youtube suggestion cards with arrows work better (although I don't quite know how to achieve that result)?


azie-ginanjar:
Hi @alexmotoc you won't get intent from wit.api because it's not been trained yet. I think @swapagarwal is in progress to add ability to mock intent #274

swapagarwal:
Hi @alexmotoc, you can now add your queries [here](https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/local/wit.json) for testing. 😄 
Can you paste a link / screenshot of what data the API provides.

alexmotoc:
This is the main data the API provides. There is way more information I couldn't capture such as the quantity of every nutrient.

screenshot from 2017-08-22 11-49-12

edadesd:
Here's a link to the API's documentation: https://developer.edamam.com/edamam-docs-recipe-api

alexmotoc:
Any updates on this?

PR #278, opened 2017-08-28
Add Trailer link

jwetzell:
Added Trailer link as a button below the IMDb link if the Trailer exists in the TMDB api call for videos.

### DIFF
--- a/modules/src/movie.py
+++ b/modules/src/movie.py
@@ -34,8 +34,8 @@ def process(input, entities):
                 'append_to_response':'videos'
             })
             data = r.json()
-        # Seperate Videos from rest of data 
-        videos = data['videos']['results']
+            # Seperate Videos from rest of data 
+            videos = data['videos']['results']
 
 	# Fetch movie rating from IMDb
         ia = IMDb()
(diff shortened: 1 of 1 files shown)


modules/src/movie.py
```diff
@@ -47,6 +50,12 @@ def process(input, entities):
         template = ButtonTemplate(text)
         template.add_web_url('IMDb Link', 'https://www.imdb.com/title/' + data['imdb_id'] + '/')
 
+	# Append first Trailer URL if one exists
+	for video in videos:
+		if video['type'] == 'Trailer':
+			template.add_web_url('Trailer Link','https://www.youtube.com/watch?v='+video['key']+'/')
```
swapagarwal: Does it always return a `YouTube` video?

swapagarwal:
Nice work @jwetzell! 
Please convert all tabs to spaces. 😅

modules/src/movie.py
jwetzell: I am actually not sure if it will always return a YouTube video but the movies I have tried did return a YouTube video I can look further in to see if there is a way to tell for sure.

jwetzell:
I have changed to tabs to spaces, let me know if anything else need to be done

swapagarwal:
Thanks @jwetzell! 👍

PR #280, opened 2017-08-29
Created food module in modules.src

finleylau:
Added to modules.src __init__.py
Added tests in test_food.py in modules.tests
Added API key in config.py


finleylau:
Some examples of testing queries:
Is Giordano's pizza in Chicago good?
What are some restaurants in San Francisco?
Chipotle in New York

finleylau:
@swapagarwal are you able to train the wit.ai so I can see if the tests pass?  Thanks!

modules/src/food.py
```diff
@@ -0,0 +1,93 @@
...
+    output = {}
+    try:
+        requested_location = entities['restaurant_location'][0]['value']
+
+        header = {'user-key': ZOMATO_API_KEY, 'Accept': 'application/json'}
+        query = requests.get('https://developers.zomato.com/api/v2.1/locations?query=' + requested_location, headers=header).json()
+        
+        location = query['location_suggestions'][0]
+
+        search_url = 'https://developers.zomato.com/api/v2.1/search?entity_id={eid}&entity_type={etype}&lat={lat}&lon={lon}&sort=rating&order=desc'.format(eid=location['entity_id'],etype=location['entity_type'],lat=location['latitude'],lon=location['longitude'])
```
horrormyth: Let's split the line please, hard to follow :)

modules/src/food.py
```diff
@@ -0,0 +1,93 @@
...
+
+ZOMATO_API_KEY = os.environ.get('ZOMATO_API_KEY', config.ZOMATO_API_KEY)
+
+def process(input,entities):
+    output = {}
+    try:
+        requested_location = entities['restaurant_location'][0]['value']
+
+        header = {'user-key': ZOMATO_API_KEY, 'Accept': 'application/json'}
+        query = requests.get('https://developers.zomato.com/api/v2.1/locations?query=' + requested_location, headers=header).json()
```
horrormyth: Are we sure that JSON conversion wouldn't fail? What happens when the connection is dropped or malformed response received. Good to have it handled :)

modules/src/food.py
```diff
@@ -0,0 +1,93 @@
...
+
+        if 'restaurant_name' in entities:
+            name = entities['restaurant_name'][0]['value']
+            search_url += '&q={name}'.format(name=name)
+
+        if 'cuisine' in entities:
+            type = entities['cuisine'][0]['value']
+            search_url += '&cuisines={type}'.format(type=type)
+
+        query = requests.get(search_url,headers=header).json()['restaurants']
```
horrormyth: Same as before, this one seems prone to failure.

.vs/ProjectSettings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "CurrentProjectSetting": null
+}
```
horrormyth: Let's start with adding the blank line at the end of each file :)  (POSIX standard)

modules/src/food.py
```diff
@@ -0,0 +1,93 @@
...
+
+        header = {'user-key': ZOMATO_API_KEY, 'Accept': 'application/json'}
+        query = requests.get('https://developers.zomato.com/api/v2.1/locations?query=' + requested_location, headers=header).json()
+        
+        location = query['location_suggestions'][0]
+
+        search_url = 'https://developers.zomato.com/api/v2.1/search?entity_id={eid}&entity_type={etype}&lat={lat}&lon={lon}&sort=rating&order=desc'.format(eid=location['entity_id'],etype=location['entity_type'],lat=location['latitude'],lon=location['longitude'])
+        
+        name = ''
+        type = ''
```
horrormyth: Let's avoid using inbuilt keywords

modules/src/food.py
```diff
@@ -0,0 +1,93 @@
...
+
+        if 'restaurant_name' in entities:
+            name = entities['restaurant_name'][0]['value']
+            search_url += '&q={name}'.format(name=name)
+
+        if 'cuisine' in entities:
+            type = entities['cuisine'][0]['value']
```
horrormyth: Same as before lets avoid using the keyword variables :)

modules/src/food.py
```diff
@@ -0,0 +1,93 @@
+import os
+import json
+import requests
+import random
+import difflib
+
+import config
+from templates.generic import *
```
horrormyth: Let's import the module that we need explicitly rather than importing everything, very dangerous :)

modules/src/food.py
```diff
@@ -0,0 +1,93 @@
...
+                
+                text = template.get_text()
+                template = ButtonTemplate(text)
+                template.add_web_url('Restaurant Link', out_rest['url'])
+                
+                output['input'] = input
+                output['output'] = template.get_message()
+                output['success'] = True
+
+    except:
```
horrormyth: This is unbearable exception, do we know what kind of exception are we catching ?

modules/src/food.py
```diff
@@ -0,0 +1,93 @@
...
+        if 'restaurant_name' in entities:
+            name = entities['restaurant_name'][0]['value']
+            search_url += '&q={name}'.format(name=name)
+
+        if 'cuisine' in entities:
+            type = entities['cuisine'][0]['value']
+            search_url += '&cuisines={type}'.format(type=type)
+
+        query = requests.get(search_url,headers=header).json()['restaurants']
+
```
horrormyth: Could you please explain what is happening here :) a bit confusing and may be I sense that it could be simplified from the line 25, And it would be better if we have a separate function with the readable name when the code is longer.

modules/src/food.py
```diff
@@ -0,0 +1,93 @@
...
+        if name == '':
```
horrormyth: Could you please explain what is happening from here :) a bit confusing and may be I sense that it could be simplified from the line 25, And it would be better if we have a separate function with the readable name when the code is longer.

modules/src/food.py
```diff
@@ -0,0 +1,93 @@
...
+    except:
+        error_message = 'I couldn\'t find any restaurants matching your query.'
+        error_message += '\nPlease ask me something else, like:'
+        error_message += '\n  - Is Giordano\'s pizza in Chicago good?'
+        error_message += '\n  - Chipotle in Evanston'
+        error_message += '\n  - What are some restaurants in Chicago?'
+        output['error_msg'] = TextTemplate(error_message).get_message()
+        output['success'] = False
+
+    return output
```
horrormyth: Let's add blank line at the end :)

modules/tests/test_food.py
```diff
@@ -0,0 +1,8 @@
+import modules
+
+
+def test_food():
+    assert ('food' == modules.process_query('Is Giordano\'s pizza in Chicago good?')[0])
+    assert ('food' == modules.process_query('Chipotle in Evanston')[0])
+    assert ('food' == modules.process_query('What are some restaurants in Chicago')[0])
+    assert ('food' != modules.process_query('something random')[0])
```
horrormyth: Let's add blank line at the end the final one :)

modules/src/food.py
finleylau: Basically, queries the api again for more specific results on a specific restaurant or type of food from the input

modules/tests/test_food.py
finleylau: I've added blank lines to the ends of the two files but github's just not displaying them in the preview windows but they're there

finleylau:
@horrormyth I fixed the variable names, the line formatting, and the blank lines at the end.  I also deleted the .vs folder (some files Visual Studio made that don't affect the code at all).  I'm not sure how to deal with the JSON conversion failures or the exception that you were talking about so I could use a little help there :)

modules/src/food.py
horrormyth: lets start with renaming the query as the response that is more readable ,and then we shall do ‘‘‘ if response.ok'''

horrormyth: lets start with renaming the query as the response that is more readable , and then we shall do 
```
if response.ok:
    content = response.json()
    try:
        location=content['location_suggestions'][0]
    except (KeyError,IndexError):
        #do what you gotta do.
    #do normal thing
```
This is still hacky way but at least we have something to catch if fails.An I hope this will make it clear :)

modules/src/food.py
horrormyth: @swapagarwal  I saw other modules code as well and try and except used seems to be very unbounded and is not helpful.And here @finleylau is trying to make a module here with the same implementation.Would you be able to look at it, please?

modules/src/food.py
finleylau: updated with your changes, is that what you were looking for?

modules/src/food.py
```diff
@@ -0,0 +1,123 @@
...
+			try:
+				location = content['location_suggstions'][0]
+			except (KeyError,IndexError):
+				error_message = 'I couldn\'t find any restaurants matching your query.'
+				error_message += '\nPlease ask me something else, like:'
+				error_message += '\n  - Is Giordano\'s pizza in Chicago good?'
+				error_message += '\n  - Chipotle in Evanston'
+				error_message += '\n  - What are some restaurants in Chicago?'
+				output['error_msg'] = TextTemplate(error_message).get_message()
+				output['success'] = False
```
horrormyth: @finleylau  This seems to be duplicated everywhere, Could we modularize the logic, please :)

modules/src/food.py
```diff
@@ -0,0 +1,123 @@
...
+													'\nAverage Price for Two: ' + out_rest['average_cost_for_two'])
+				
+							text = template.get_text()
+							template = ButtonTemplate(text)
+							template.add_web_url('Restaurant Link', out_rest['url'])
+				
+							output['input'] = input
+							output['output'] = template.get_message()
+							output['success'] = True
+				except (KeyError,IndexError):
```
horrormyth: @finleylau Do you think we will be getting KeyyErro and Index Error here? Because it seems that you have a chunk of code in the try block. Let's modularize, please :)

finleylau: Can you please further explain?  Sorry, I'm not sure what you mean!

modules/src/food.py
finleylau: done!

PR #283, opened 2017-09-09
Added cryptocurrency feature

shawnjzlee:


modules/src/crypto.py
```diff
@@ -0,0 +1,41 @@
+import requests
+
+from templates.text import TextTemplate
+
+
+def process(input, entities):
```
horrormyth: It would be nice if we do not use inbuilt keywords and function as a variable :)

modules/tests/test_crypto.py
```diff
@@ -0,0 +1,26 @@
...
+    intent, entities = modules.process_query('market ETH to OMG')
+    assert ('crypto' == intent)
+    assert ('ETH to OMG' == entities['market'][0]['value'])
+
+    intent, entities = modules.process_query('crypto ETH to OMG')
+    assert ('crypto' == intent)
+    assert ('ETH to OMG' == entities['market'][0]['value'])
+
+    assert ('crypto' != modules.process_query('tests invalid input')[0])
+
```
horrormyth: Let's add a bank line :)

modules/src/crypto.py
```diff
@@ -0,0 +1,41 @@
...
+
+        if data['success'] == False:
+            market += '-' + market[:market.find('-')]
+            market = market[market.find('-')+1:]
+            r = requests.get('https://bittrex.com/api/v1.1/public/getticker?market=' + market)
+            data = r.json()
+
+        bid = data['result']['Bid']
+        ask = data['result']['Ask']
+        last = data['result']['Last']
```
horrormyth: Lets put these in try and except.

modules/src/crypto.py
shawnjzlee: @horrormyth would `entry` be a good replacement for `input`? (:

shawnjzlee:
I apologize for the commit messages.

@horrormyth can you please check if i correctly implemented the try and except correctly? Thanks!

modules/src/crypto.py
```diff
@@ -3,39 +3,52 @@
...
+        output['error_msg'] = error_message
+        output['success'] = False
+
     try:
         market = entities['market'][0]['value'].upper()
         if 'to' in market:
-            market.replace('to', '-').replace(' ', '')
+            market = market.replace('to', '-').replace(' ', '')
 
         r = requests.get('https://bittrex.com/api/v1.1/public/getticker?market=' + market)
```
horrormyth: Let's verify if the response is okay :) and then we could do the JSON loading.

modules/src/crypto.py
```diff
@@ -3,39 +3,52 @@
...
         data = r.json()
 
-        if data['success'] == False:
+        if not data['success']:
             market += '-' + market[:market.find('-')]
             market = market[market.find('-')+1:]
             r = requests.get('https://bittrex.com/api/v1.1/public/getticker?market=' + market)
```
horrormyth: same as above :)

modules/src/crypto.py
horrormyth: user_input/entry would be fine, it is always better to have the relevant variable name so that it is very readable so to track the code. :)

modules/src/crypto.py
```diff
@@ -0,0 +1,54 @@
...
+        data = r.json()
+
+        if not data['success']:
+            market += '-' + market[:market.find('-')]
+            market = market[market.find('-')+1:]
+            r = requests.get('https://bittrex.com/api/v1.1/public/getticker?market=' + market)
+            data = r.json()
+
+        if not data['success']:
+            raise BittrexAPIException(data)
```
horrormyth: [Here](https://stackoverflow.com/questions/2052390/manually-raising-throwing-an-exception-in-python)  very thoroughly described raising an exception

shawnjzlee: Thank you for the link and the guidance!

As for the question, I was following the documentation for errors and exceptions. I was planning to add a user-defined exception after I had other types of exceptions for other API calls.

PR #284, opened 2017-09-12
create cat module

mackopes:
Everyone loves cats, so I created a basic module that should return random picture of cat (using http://thecatapi.com). 
Jarvis just needs to understand command "cat" (or anything else that should return cat image). 
Right now api returns only images from limited repository of images (1000 pcs). I can apply for API key and get even bigger repo of cat images.


MatkoUSKO:
Is this "improvement" only ignored? :(

MatkoUSKO:
@swapagarwal  Could you please review the code and comment on changes that are required? Or if this is even wanted "improvement"

modules/src/cat.py
```diff
@@ -0,0 +1,48 @@
+import requests
+
+import modules
+from templates.attachment import AttachmentTemplate
+from templates.quick_replies import add_quick_reply
+from templates.text import TextTemplate
+
+def process(input, entities=None):
```
ab9-er: `input` is a built in keyword.

modules/src/cat.py
```diff
@@ -0,0 +1,48 @@
...
+        for r in response:
+            x = r.find('<url>')
+            if x != -1:
+                cat_image = r[x+5:r.find('</url')].strip(' ')
+
+
+        message = AttachmentTemplate(cat_image, type='image').get_message()
+        message = add_quick_reply(message, 'More Cats', modules.generate_postback('cat'))
+        output = {
+            'input': input,
```
ab9-er: idem.

modules/src/cat.py
```diff
@@ -0,0 +1,48 @@
...
+        #   </data>
+        # </response>
+
+        for r in response:
+            x = r.find('<url>')
+            if x != -1:
+                cat_image = r[x+5:r.find('</url')].strip(' ')
+
+
+        message = AttachmentTemplate(cat_image, type='image').get_message()
```
ab9-er: `type` is a built in keyword.

modules/src/cat.py
```diff
@@ -0,0 +1,48 @@
...
+
+import modules
+from templates.attachment import AttachmentTemplate
+from templates.quick_replies import add_quick_reply
+from templates.text import TextTemplate
+
+def process(input, entities=None):
+    output = {}
+    try:
+        response = requests.get('http://thecatapi.com/api/images/get', params = {'format': 'xml'})
```
ab9-er: No spaces around `=` in args, kwargs expressions. Ref PEP guides.

modules/src/cat.py
```diff
@@ -0,0 +1,48 @@
...
+            'input': input,
+            'output': message,
+            'success': True
+        }
+    except:
+        error_message = 'There was some error while retrieving your cat picture.'
+        output = {
+            'error_msg': TextTemplate(error_message).get_message(),
+            'success': False
+        }
```
ab9-er: Why doesn't the `output` object have the same structure for both success and error?

MatkoUSKO:
@ab9-er I basically followed the form of all other modules I found. And I wanted to maintain the same style that was used before. Same goes for the structure of output objects in case of success/error. Also, keyword "type" is specified in AttachmentTemplate.

ab9-er:
@swapagarwal must change the `type` keyword.

ab9-er:
Ideally, success and error having same structure will help in the maintaining the response parser and helps also to extend the module.

PR #285, opened 2017-09-15
Add random wallpaper module

conormag94:
I've created a new module to allow users to get a random photo from [Unsplash](https://unsplash.com/) to use as a wallpaper for their phones/desktops. People could use the photos for whatever they like though - wallpapers were just what I had in mind initially. I also thought of expanding the module to allow to search wallpapers but I didn't in case a newcomer wanted to tackle it at some stage.

Some example queries:
- "I need a new wallpaper"
- "wallpaper"
- "Show me a random wallpaper"
- etc.

I mocked the wit intent locally and the module seems to be working fine. I was just a bit unsure of which information to put in the template response so that can always be modified if needed.

**You'll need to create a developer account [here](https://unsplash.com/developers) and register JARVIS to get an API key (Application-ID).**


horrormyth:
@conormag94  Lets fix the test please :),  one of the tests is failing.

conormag94:
@horrormyth Yeah I believe that this test is failing because Jarvis doesn't know how to process the new module. 

@swapagarwal will need to add training in wit.ai when he gets a chance. Alternatively I could add some sample intent in the wit json file but my impression was that this was for local debugging only so I took it out of the final PR.

conormag94:
Hi @swapagarwal did you get a chance to have a look at this PR?

PR #287, opened 2017-09-27
Fix typo in introduction to user guide 

isbee:
'die' does not working, so i changed it to 'dice'.


djung460:
Not sure how the devs want commit messages to be formatted but this is always a good call https://chris.beams.io/posts/git-commit/

isbee:
@djung460  Thanks for the advice!

swapagarwal:
This has now been fixed.

PR #288, opened 2017-09-27
Fix a bug for "roll a die" query

iamblizzard:
Resolve this issue by inserting query for the same in local testing file (local/wit.json) and creating die.py in (modules/src/).


modules/__init__.py
```diff
@@ -47,16 +49,24 @@ def search(input, sender=None, postback=False):
...
             r = requests.get('https://graph.facebook.com/v2.6/' + str(sender), params={
                 'fields': 'first_name',
                 'access_token': os.environ.get('ACCESS_TOKEN', config.ACCESS_TOKEN)
             })
+
             if entities is None:
                 entities = {}
             entities['sender'] = r.json()
+
+        if intent == 'die':
```
djung460: I suggest not putting this here, adding this check sets a precedent that we should check all intents and then match it to their correct module. @swapagarwal what do you think about creating a function that checks a lookup table to map intent to their respective module?

modules/src/die.py
```diff
@@ -0,0 +1,28 @@
+import random
```
djung460: This file should be removed. There is already a dice.py that is the exact same.

swapagarwal:
Hey @iamblizzard, this has now been fixed! See the commit history for reference. 😄

PR #292, opened 2017-10-03
wiki: Added random page as quick reply 

aviral1701:
I have tried adding a quick reply for wiki.Please review

### DIFF
--- a/modules/src/wiki.py
+++ b/modules/src/wiki.py
@@ -2,6 +2,7 @@ import wikipedia
 
 from templates.generic import *
 from templates.text import TextTemplate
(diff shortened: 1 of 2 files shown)


modules/src/wiki.py
```diff
@@ -14,7 +15,10 @@ def process(input, entities):
         text = template.get_text()
         template = ButtonTemplate(text)
         template.add_web_url('Wikipedia Link', data.url)
-        output['output'] = template.get_message()
+        page = wikipedia.random(pages=1)
```
swapagarwal: This variable seems unused. 😅

PR #295, opened 2017-10-04
Added Docker configuration for local setup

### LINKED ISSUE #291: Add docker support

TheSmokingGnu:
Added Dockerfile with minimal configuration and updated the readme with instruction on how to use

Fixes #291

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,45 @@
+####################################################
+# Use python 2.x and the latest version of alpine  #
+####################################################
+FROM python:2-alpine
+
+LABEL maintainer=peter@grainger.xyz
+
+###################################################
+# Add a compiler                                  #
+###################################################
+RUN apk update && \
+    apk add --no-cache gcc \
+                       libffi-dev \
+                       openssl-dev \
+                       libxslt-dev \
+                       libxml2-dev \
+                       musl-dev \
+                       linux-headers
+
+###################################################
+# Create somewhere to put the files               #
+###################################################
+RUN mkdir -p /opt/jarvis
+WORKDIR /opt/jarvis
+
+###################################################
+# Requirements file first to help cache           #
+###################################################
+COPY requirements.txt .
+RUN pip install -r requirements.txt
+
+###################################################
+# Copy over the rest.                             #
+###################################################
+COPY ./ ./
+
+###################################################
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+###################################################
+# Create somewhere to put the files               #
+###################################################
+RUN mkdir -p /opt/jarvis
+WORKDIR /opt/jarvis
+
+###################################################
+# Requirements file first to help cache           #
+###################################################
+COPY requirements.txt .
```
ab9-er: You should ideally always place your requirements before the code base. This will prevent re-installing the requirements if a small part of the source code changes. Doesn't make a lot of difference but saves plenty of build time.

ab9-er: Sorry I didnt pay attention to the `COPY ./ ./`, my bad 😞

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+###################################################
+COPY ./ ./
+
+###################################################
+# Expose the port externally from the container   #
+###################################################
+EXPOSE 5000
+
+###################################################
+# Run jarvis                                         #
```
swapagarwal: Please indent the last `#`.

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+###################################################
+CMD python jarvis.py
```
swapagarwal: Add a new line here.

Dockerfile
```diff
@@ -0,0 +1,45 @@
+####################################################
+# Use python 2.x and the latest version of alpine  #
+####################################################
+FROM python:2-alpine
+
+LABEL maintainer=peter@grainger.xyz
```
swapagarwal: Let me know if you would you be willing to maintain this. 😄

Dockerfile
swapagarwal: Pressing `Ctrl+C` doesn't exit it. I had to run `docker ps`, then `docker stop <id>` to terminate it.
Changing it to CMD ["python", "jarvis.py"] seemed to work as expected!

swapagarwal:
Thanks @ab9-er for the insight.
Great work @TheSmokingGnu! The docker image is almost in place. 🎉

PR #298, opened 2017-10-05
Video: Fix query breaking on a channel result

### LINKED ISSUE #297: video link in persistent menu not working
In the persistent menu, go to "Common Queries", "Media", and click on "videos of sia".
It returns an error.
Investigate why this is not working.

oddurad:
Fixes #297

* Update video.py

There appears to be an issue with the YouTube Data API that is causing
it to return channel results from the /search API. Added a filter to
the returned API data to filter out all non-video results.

See: https://stackoverflow.com/questions/45527041/youtube-api-video-only-search-returning-a-channel-object

The above issue is only encountered on some queries and not all, for example "sia videos" would return a channel result while "eminem videos" would not.


swapagarwal:
1 down, 3 to go.. https://hacktoberfestchecker.herokuapp.com/?username=oddurad

PR #300, opened 2017-10-08
Added Urban Dictionary module

xdavidhu:


xdavidhu:
@swapagarwal Please add this new module to witt.ai beacuse otherwise the test will not succeed.

I tried to fix it but ended up reverting back to the original commit which should work if you set witt.ai up.

PR #302, opened 2017-10-12
Upper case 'Sir' bux fixed & Better name handling in hello module

xdavidhu:
The Sir with the capital s was not replaced when the user had a name in the entities object.

### DIFF
--- a/modules/src/hello.py
+++ b/modules/src/hello.py
@@ -5,12 +5,12 @@ from templates.text import TextTemplate
 
 def process(input, entities=None):
     greetings = [
-        'Welcome home, {0}...',
-        'All wrapped up here, {0}. Will there be anything else?',
-        '{0}, I think I need to sleep now...',
+        'Welcome home, sir...',
+        'All wrapped up here, sir. Will there be anything else?',
+        'Sir, I think I need to sleep now...',
         'I seem to do quite well for a stretch, and then at the end of the sentence I say the wrong cranberry.',
-        'At your service, {0}.',
-        'Oh hello, {0}!',
+        'At your service, sir.',
+        'Oh hello, sir!',
         'Perhaps, if you intend to visit other planets, we should improve the exosystems.',
     ]
 
@@ -21,13 +21,8 @@ def process(input, entities=None):
         if 'sender' in entities and 'first_name' in entities['sender']:
             personalized = True
             sender_name = entities['sender']['first_name']
-            respone = respone.format(sender_name)
-
-    if not personalized:
-        if respone[:3] == "{0}":
-            respone = respone.format("Sir")
-        else:
-            respone = respone.format("sir")
+            respone = response.replace('Sir', sender_name)
+            respone = response.replace('sir', sender_name)
 
     output = {
         'input': input,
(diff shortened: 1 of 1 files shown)


swapagarwal:
Nice catch! 💯

modules/src/hello.py
```diff
@@ -5,21 +5,33 @@
...
+        '{0}, I think I need to sleep now...',
         'I seem to do quite well for a stretch, and then at the end of the sentence I say the wrong cranberry.',
-        'At your service, sir.',
-        'Oh hello, sir!',
+        'At your service, {0}.',
+        'Oh hello, {0}!',
         'Perhaps, if you intend to visit other planets, we should improve the exosystems.',
     ]
+
+    respone = random.choice(greetings)
```
swapagarwal: response*

modules/src/hello.py
```diff
@@ -5,21 +5,33 @@
...
+    personalized = False
     if entities is not None:
         if 'sender' in entities and 'first_name' in entities['sender']:
+            personalized = True
             sender_name = entities['sender']['first_name']
-            greetings = [greeting.replace('sir', sender_name) for greeting in greetings]
+            respone = respone.format(sender_name)
+
+    if not personalized:
+        if respone[:3] == "{0}":
```
swapagarwal: Magic String! This looks a bit hacky. Can we do something better here? 😄

xdavidhu: That's not a magic string, its used for the format function to add new text into the string. Not sure what to change since I have to check if its at the start or not because it has to decide that the 'S' is capital or not.

swapagarwal: Sorry, my bad! I meant magic number (referring to 3).

xdavidhu: No problem. With the 3, I'm basically just checking if the name is the first word or not, so when it is, the first 3 char is "{0}" then its a "Sir" with a capital because its the start of the sentence.  You see, the 3 can be fixed without any problems because the name always will be "{0}", a 3 char long string.

swapagarwal: A better way would be to replace sir using a case-insensitive method. That'll require a minimal change. What do you think? 😇

xdavidhu:
I added the new case insensitive replace.

modules/src/hello.py
```diff
@@ -13,13 +13,20 @@ def process(input, entities=None):
         'Oh hello, sir!',
         'Perhaps, if you intend to visit other planets, we should improve the exosystems.',
     ]
+
+    respone = random.choice(greetings)
+
+    personalized = False
```
swapagarwal: unused variable

modules/src/hello.py
```diff
@@ -13,13 +13,20 @@ def process(input, entities=None):
...
     if entities is not None:
         if 'sender' in entities and 'first_name' in entities['sender']:
+            personalized = True
             sender_name = entities['sender']['first_name']
-            greetings = [greeting.replace('sir', sender_name) for greeting in greetings]
```
swapagarwal: I meant we can just change this single line to have the replacement case-insensitive. 😄 
What do you think?

ab9-er: `re`!

PR #325, opened 2017-10-22
Movie: Add trailer

ed-word:
Fix Movie: Add trailer link #259


swapagarwal:
Hi @ed-word, thanks for your contribution but this is already being worked upon in #278
Feel free to take up any other issues. 😄

PR #332, opened 2017-10-27
What? Cocktail?

jaskaransarkaria:
Sorry my previous pull request was an accident and came from a branch which i had not yet merged into my own up to date master.


ab9-er:
Squash wouldn't hurt!

modules/src/what_cocktail.py
```diff
@@ -0,0 +1,351 @@
+#rewrite the program (which a burglar stole from me!) and then release it onto telegram and messenger!
+''' argh what cocktail can i make with these ingredients? search and return a cocktail which has
+the most matching ingredients'''
+from enchant.checker import SpellChecker
+import re
+
+print("What? Cocktail")
```
ab9-er: This is more like a Python3 syntax for print.

modules/src/what_cocktail.py
```diff
@@ -0,0 +1,351 @@
...
+print()
+print()
```
ab9-er: Empty print?

modules/src/what_cocktail.py
```diff
@@ -0,0 +1,351 @@
...
+
+def search(drinks, dictionary):
+    filter_dict = {}
+    for key, value in dictionary.items():
+        for v in value:
+            if v == drinks:
+                filter_dict.update({key: value})
+    drinks = filter_dict
```
ab9-er: You could directly return `filter_dict` instead of assigning it to drinks. It has no impact on the code.

modules/src/what_cocktail.py
```diff
@@ -0,0 +1,351 @@
...
+            print("Or...")
+        elif last_opt == 1:
+            print()
+            print(key)
+            print("made with these measures:")
+            print(final_print(value))
+            print()
+            exit()
+
+cocktails = {
```
ab9-er: Would appreciate putting it either on the top of the file. Better to put it in a `.json` file and parse it in the script.

modules/src/what_cocktail.py
```diff
@@ -0,0 +1,351 @@
...
+    if dict_lookup(word, cocktails) is True:
+        return word
+
+    elif word == "Rum":
+        word = input("White or Dark?").title()
+        word += " Rum"
+        chkr = SpellChecker("en_GB", word)
+        for err in chkr:
+            while dict_lookup(word, cocktails) is not True:
+                print("Your spelling wasn't recognised")
```
ab9-er: idem

modules/src/what_cocktail.py
```diff
@@ -0,0 +1,351 @@
...
+                suggestion = str(err.suggest(word))
+                suggestion = re.sub(r'(\')', '', suggestion)
+                suggestion = re.sub(r'\[', '', suggestion)
+                suggestion = re.sub(r'\]', '', suggestion)
+                print(suggestion)
+                print()
+                word = input("Please re-type it:").title()
+                print()
+                if word == "White":
+                    word = "White Rum"
```
ab9-er: return word instead of assigning it.

modules/src/what_cocktail.py
jaskaransarkaria: it is python3, is that an issue?

modules/src/what_cocktail.py
jaskaransarkaria: I inserted these to make the output more readable, I can replace them with print("\n") if you would prefer?

modules/src/what_cocktail.py
jaskaransarkaria: done

modules/src/what_cocktail.py
ab9-er: Yep, that would be more readable - adding `\n` at the end of the string.

modules/src/what_cocktail.py
ab9-er: 👌

modules/src/what_cocktail.py
jaskaransarkaria: i've put it on the top of the file for time being, i'm a beginner programmer and have not yet got to grips with .json yet

modules/src/what_cocktail.py
jaskaransarkaria: done

modules/src/what_cocktail.py
jaskaransarkaria: sorry not sure what you mean here. I do return it but I also assign it. What would you like me to change?

modules/src/what_cocktail.py
jaskaransarkaria: Ahhhh I didn't realise!! hmm do you suggest rewriting for python 2 or is there the possibility of a python3 version coming out? or will it just work?

modules/src/what_cocktail.py
jaskaransarkaria: done

modules/src/what_cocktail.py
jaskaransarkaria: also who do I update you with my file changes?

modules/src/what_cocktail.py
```diff
@@ -0,0 +1,351 @@
...
+    statement2 = re.sub(r'\s,', ',', statement2)
+    statement2 = re.sub(r'\s$', '', statement2)
+    statement2 = re.sub(r'\s\s', ', ', statement2)
+    print("%s made with %s." %(statement, statement2))
+
+def dict_lookup(ing, dict):
+    for key, value in dict.items():
+        for v in value:
+            if v == ing:
+                return True
```
ab9-er: I have a neat little snippet to change this one, since you never use `key` why bother put it in memory.

```
return ing in [item for sublist in list(dict.values()) for item in sublist]
```

modules/src/what_cocktail.py
```diff
@@ -0,0 +1,351 @@
...
+    statement2 = re.sub(r'a?A?\d+\sS?s?plashe?s?', '', statement2)
+    statement2 = re.sub(r'T?t?o\sG?g?arnish', '', statement2)
+    statement2 = re.sub(r'T?t?o\sT?t?op\sU?u?p', '', statement2)
+    statement2 = re.sub(r'\d+.', '', statement2)
+    statement2 = re.sub(r'\s,', ',', statement2)
+    statement2 = re.sub(r'\s$', '', statement2)
+    statement2 = re.sub(r'\s\s', ', ', statement2)
+    print("%s made with %s." %(statement, statement2))
+
+def dict_lookup(ing, dict):
```
ab9-er: Also, please change the variable name from `dict` to something else. It is a Python reserved keyword.

modules/src/what_cocktail.py
ab9-er: You don't use the variable `word` so why bother assigning it? Instead do `return "White Rum"` or `return "Dark Rum"` instead. Or assign it to `word` and have just one return at the end of the function

modules/src/what_cocktail.py
ab9-er: I guess the update is something only @swapagarwal has a clue about. But stick to Py2 for the moment.

modules/src/what_cocktail.py
ab9-er: Small modules would be easy to maintain. I suggest having a look at https://stackoverflow.com/questions/20199126/reading-json-from-a-file

PR #335, opened 2017-11-08
Update README.md

dkansk1323:
Add Example in 'Local Development / Testing'


swapagarwal:
@dkansk1323 Can you add the response as well? Maybe a permalink

dkansk1323:
@swapagarwal  I don't know why my commits cannot pass the test...

PR #342, opened 2017-11-16
Move from hello.py to bye.py

dkansk1323:
Move 'Sir, I think I need to sleep now...' from hello to bye

PR #346, opened 2017-11-18
Deleted inadequate response for 'hello'

orbit-stabilizer:
Specifically, this line was deleted:         
'Sir, I think I need to sleep now...',
It will be moved to the response for 'bye'.


swapagarwal:

swapagarwal:
Resolved in #342

Issue #337, opened 2017-11-14
Generalize location entities


swapagarwal:
As pointed out in #165, we use `weather_location` and `time_location` in modules, but they hold similar type of data. We can combine them to a single entity `location`.

gibertoni:
Hi, I am new to the project but since this is a low hanging fruit, I may work on this.
Is it ok for you?

DanielDo98:
Is this still open? If so, could I take a crack at it?

msultani:
@DanielDo98 have you worked on this? If not, I'd like to give it a shot.

chrisfoose:
Can I give this a shot?  I really don't know what I'm doing but I suppose I can figure it out.

brandonlbarrow:
Opened a PR for this - #794

Cerabbite:
Is this issue still open? If so, can I try to solve it?

Alrwx:
Hey is this issue still open? I'm gonna start taking a crack at it!

PR #348, opened 2017-11-18
Abstracting location entity as requested by issue #337

gibertoni:
Adding a location module to abstract the use of the Google Maps API (geocode)
Is this the desired approach to issue #337?

Also, my editor follows PEP08 guideline and I changed some formatting on the files I edited.
Please tell me if this is not desired or if the code is less readable this way.

### DIFF
--- a/modules/src/time.py
+++ b/modules/src/time.py
@@ -1,30 +1,34 @@
 import os
-from datetime import datetime
-
 import requests
-
 import config
-from templates.text import TextTemplate
 
-MAPQUEST_CONSUMER_KEY = os.environ.get('MAPQUEST_CONSUMER_KEY', config.MAPQUEST_CONSUMER_KEY)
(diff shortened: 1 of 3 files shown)


swapagarwal:

swapagarwal:
Hey @gibertoni, this task just required merging the entities, i.e. renaming `weather_location` and `time_location` to `location`. 😄

gibertoni:
I see.
Since the entity is something we created on wit.ai, shouldn't it be edited there?
I don't understand how I handle it on the jarvis/module code.
Can you point me to a starting point?

gibertoni:
I see. I think I understand it now.
So I should update the time.py and weather.py modules to expect an entity called 'location'
And then, you will update wit.ai to return the entity with the value 'location' for those cases. 
Is that right?

PR #352, opened 2017-11-21
Prices of stocks

maulikjs:


swapagarwal:

modules/src/stocks.py
```diff
@@ -0,0 +1,28 @@
+import requests
+import json
+
+from templates.text import TextTemplate
+
+
+def process(input, entities):
```
ab9-er: Can you use any other variable name instead of `input`. It is a Python built-in https://docs.python.org/2/library/functions.html#input

maulikjs: gotcha, on it.
Any ideal how to train wit.ai for JARVIS?

maulikjs: and pass the tests?

maulikjs: @ab9-er Any ideal how to train wit.ai for JARVIS and pass the tests?

modules/src/stocks.py
```diff
@@ -0,0 +1,28 @@
+import requests
+import json
+
+from templates.text import TextTemplate
```
ab9-er: Just curious to know the reason you chose `templates`

modules/src/stocks.py
```diff
@@ -0,0 +1,28 @@
...
+
+        r = requests.get('http://www.google.com/finance?q='+stock_name+'&output=json')
+        data = json.loads(r.text[3:])
+        stockprice = data[0]['l']
+
+        returnString = 'The last price of ' + stock_name + ' is: USD '+ stockprice
+        output['input'] = userInput
+        output['output'] = TextTemplate(returnString).get_message()
+        output['success'] = True
+    except:
```
ab9-er: Would be a lot better if you could specify the exceptions. This statement will catch a broad variety of exceptions which won't be something that the script would want.

maulikjs: I based my code on another module I just assumed we are doing a generic error message for each module. But sure i can make them more specific based on whats happening.

ab9-er:
@maulikjs never worked on `wit.ai` but I sure can have a look at it and get back to you.

ab9-er:
Now that I see the tests that you have written, they seem more to me like integration level tests than basic unit tests. Why don't you try stubbing the service instead?

ab9-er:
@maulikjs Take a look at these tests here https://github.com/lextoumbourou/PyWit/blob/master/tests/unit/test_wit.py

maulikjs:
Gotcha. The thing is in offline testing with me adding the intent and values in wit.json all the tests pass and things work well. What I feel is going wrong here is wit.ai doesn't recognize that it needs to read stocks.  But if the apps Wit.ai (which I assume is under @swapagarwal  account). And if that thing was done then this should work fine along with passing all the tests. @swapagarwal what are you thoughts on this?
I can stub the service but that would just be to pass the tests. The module still wont work.

ab9-er:
Do you mean your tests will reach out to wit.ai to get the results ?

PR #355, opened 2017-11-22
Add love module

ghdud4006:
eg)
user: i love you
jarvis: I love you like crazy.

user: do you love me?
jarvis: I'm all yours.

As above, i added module that express love.


swapagarwal:

Bilalharoon:
i dont see the point

but okay

PR #391, opened 2018-03-22
Add Maze Module

merlinlikethewizard:

A new maze module, maze.py!

Allows user to ask for a maze of a certain size, and responds with a randomly generated maze of the requested size, with only one solution. The returned maze contains unicode lines, corners, etc., as well as new line characters.

process() is set up to take the maze size from entities['size'][0]['value'], but that can change.


swapagarwal:

swapagarwal:
I don't see any maze.py file [here](https://github.com/swapagarwal/JARVIS-on-Messenger/pull/391/files).
By the way, what's the intended use-case of this module?

PR #396, opened 2018-03-31
Bmi module

barbarkahoufkova:
Bmi module can serve as a tool to evaluate person's health status by calculating its body mass index. 
BMI can be measured in kilograms and centimeters or in pounds and inches. For the purpose of the program, the first option is called metric and the second option is called imperial. BMI is either a weight in pounds divided by the square of height in inches, all multiplied by 703 or a weight in kilograms divided by the square of height in meters. The function accepts height and weight and it returns your body mass index including your category.

-
-
-

**Fixes**: #

### DIFF
--- a/modules/src/bmi.py
+++ b/modules/src/bmi.py
@@ -1,13 +1,46 @@
-
 import config
 import modules
 from templates.text import TextTemplate
+
+def bmi(string):
+	x = string.split()
+	mode = x[2]
+    height = float(x[0])
+    weight = float(x[1])
+	if mode == "imperial":
+		height_sq  = height*height
+		bmi = weight / height_sq * 703
+		bmi_final = int(round(bmi,0))
+		if bmi_final <= 18.5:
+			return ("Your BMI is {} and you belong to category underweight.".format(bmi_final))
+		elif bmi_final < 25:
+			return ("Your BMI is {} and you belong to category normal.".format(bmi_final))
+		elif bmi_final < 30:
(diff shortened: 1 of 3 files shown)


swapagarwal:

modules/src/bmi.py
```diff
@@ -0,0 +1,54 @@
+import config
+import modules
+from templates.text import TextTemplate
+
+def bmi(string):
+	x = string.split()
+	mode = x[2]
+    height = float(x[0])
```
swapagarwal: indentation does not match with above line

modules/src/bmi.py
```diff
@@ -0,0 +1,54 @@
...
+		elif bmi_final < 30:
+			return ("Your BMI is {} and you belong to category overweight.".format(bmi_final))
+		else: 
+			return ("Your BMI is {} and you belong to category obese.".format(bmi_final))
+	else:
+		heigh_new = heigh/100
+		height_sq  = height*height
+		bmi = weight / height_sq
+		bmi_final = int(round(bmi,0))
+		if bmi_final <= 18.5:
```
swapagarwal: This looks duplicated. Follow [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

modules/src/bmi.py
```diff
@@ -0,0 +1,54 @@
+import config
+import modules
+from templates.text import TextTemplate
+
+def bmi(string):
```
swapagarwal: Let's write some tests for this function.

PR #427, opened 2018-06-03
Update __init__.py

sunminxxi:
I added a module named love. This module is the module that is used when sending a message i love you.


sunminxxi:
Leave a comment

PR #428, opened 2018-06-03
Update wit.json : "i love you"

sunminxxi:
I want to create a new module called love.
The reason is that I sent a message to JARVIS that i love you, but I did not get the right answer.
The word i love you means to believe in the other person and at the same time is pleasant.
If you do not answer the most basic word i love you, JARVIS will be disappointed.
I wanted to make it even closer between users and JARVIS by adding new modules.
This will increase the credibility of the JARVIS messenger itself.

### DIFF
--- a/local/wit.json
+++ b/local/wit.json
@@ -12,5 +12,10 @@
 	"test": {
 		"intent": "xkcd",
 		"entities": []
+	},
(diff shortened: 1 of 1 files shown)


swapagarwal:

PR #515, opened 2018-12-06
Added method to return the top search result from youtube

CiaranFahy:
Added method to return the top search result from youtube.


swapagarwal:

PR #537, opened 2019-09-21
Now runs with python3 (needs more testing)

PredatorFeesh:
This should now actually run on python3, ALTHOUGH IT DOES REQUIRE TESTING AND UPDATES.

- Changed imports to work with Python3
- Checked to make sure new imports have the same function.
- Ran on `python3 jarvis.py` after installing all modules with `pip3 install -3 requirements.txt`

** Need some help updating and testing further, as so far only imports have been updated to work. **

### DIFF
--- a/modules/__init__.py
+++ b/modules/__init__.py
@@ -7,7 +7,7 @@ import requests
 
 import config
 import modules
-from src import *
+from .src import *
 from templates.quick_replies import add_quick_reply
 from templates.text import TextTemplate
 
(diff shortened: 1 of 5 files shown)


PredatorFeesh:
Didn't notice someone already started this [here!](https://github.com/swapagarwal/JARVIS-on-Messenger/pull/502) The code is very similar in both, but the required.txt file is better suited there and will be implemented here. Credit to [skelmdev](https://github.com/skelmdev) for that.

Issue #559, opened 2019-10-05
Hacktoberfest 2019


swapagarwal:
Hello everyone! 👋
JARVIS is open for Hacktoberfest 2019! 🎉

Here's an informative, beginner-friendly and FUN video on how to get started on your open-source journey: https://www.youtube.com/watch?v=jUxy4WNdc0s
@znck and @tanaypratap walk you through an end-to-end journey of contributing to JARVIS.

Here are some things that the project would benefit from your contributions:
1. Add facts/jokes/motivations/quotes here: https://github.com/swapagarwal/JARVIS-on-Messenger/tree/master/data
Keep the quality high. It's better to send one good high-quality submission than ten low-quality ones.
1. Try out the bot at https://m.me/J.A.R.V.I.S.on.Messenger
If some query does not work as expected, you can add it here: https://github.com/swapagarwal/JARVIS-on-Messenger/blob/master/local/wit.json
1. For bigger changes, open an issue for discussion before sending a PR.
1. The discord bot is in development here: https://github.com/the-vision/jarvis-discord
1. Other platforms will be added soon at https://github.com/the-vision

I might be late to merge your PR given that I have 100+ PRs to review already across my other projects, but rest assured I'll take a look as soon as possible. Note that your PR does not have to be merged to qualify for Hacktoberfest.

If you're absolutely new to the world of open-source, here's a blog post answering some of the frequently asked questions: https://medium.com/@swap/hacktoberfest-2019-time-to-start-your-open-source-journey-8227e10fac33
It also contains links of some other beginner-friendly projects to contribute in Hacktoberfest.

To iterate, aim for quality over quantity and enjoy the ride! ✌️

ahadmushir:
Hey. 
I would like to contribute for facts and jokes section on behalf of Hacktoberfest :)

prathameshThakur:
Added some interesting facts! #824

adam-weiler:
I added a fact, joke, motivation, and quote. #915

PR #579, opened 2019-10-05
Patch 2

yogesh-hackx:


yogesh-hackx:
[swapagarwal#559](https://github.com/swapagarwal/JARVIS-on-Messenger/issues/559)

PR #655, opened 2019-10-08
Never tell the same joke twice

halversonmd:
Current set of jokes is limited, making it likely to tell the same joke twice.
Each time Jarvis tells a joke, store in "used_jokes" and download a new joke from `https://icanhazdadjoke.com`

-
-
-

**Fixes**: #


swapagarwal:
I don't understand the used_jokes thing. The bot is used by multiple people simultaneously. How would it work?

PR #661, opened 2019-10-09
Added fact

ghost:
Added a fact about bobcats.


ghost:
Added a joke as well.

Issue #414, opened 2018-05-27
Add new facts to fact.py


limmiok:
I want to add a new facts to fact.py!!!!
How about this facts?
"In 1930 the Wright brothers flew for 59 seconds." 
"38 years later the Japanese bombed Pearl Harbor."
"28 years after that, we landed on the moon.”

Please check!!!!!

mansoorameen:
Hey @swapagarwal ,can you add hacktoberfest label to this?

PR #672, opened 2019-10-09
Add fact earth's surface.

thakur24:
Added a fact about earth's surface.

It adds a fact in #414

-Adding an interesting fact regarding Earth's surface which is good to know .
-
-

**Fixes**: #


thakur24:
Added an interesting fact !

Issue #482, opened 2018-06-04
Currency Conversion is not working


ftang1996:
typing "usd to eur" does not work and neither does hitting the menu option for this

Jarvis responds with 
"I couldn't convert between those currencies.
Please ask me something else, like:
  - HKD to USD
  - USD to EUR rate
  - how much is 100 USD to INR"

ftang1996:
I would like to see if i can fix this issue if that's ok

ChristopherDaw:
Currency conversion is currently broken because the api request to `api.fixer.io` is invalid.  I looked into it and the free api access only gives 1,000 api requests per month.

A possible way around the limit per month would be scraping google after a conversion search.  I'm not sure of the legality of scraping Google for the currency conversion rate instead of sending an api request, however.

illichosky:
What's the status on this? I'd love to contribute to fix it

ChristopherDaw:
@illichosky Take a crack at it! I just diagnosed the issue, I wasn't planning on fixing it :+1:

illichosky:
@ChristopherDaw Hey Chris! I'd love to take this issue. I'm super busy this week, but I'm planning to PR a fix by next week. Could you assign me to the issue?

ChristopherDaw:
@illichosky I'm not in charge here and I didn't open the issue so I don't think I can assign you to it. When you're done you can always reference the issue from the PR.

bowling1:
Perhaps the API used to pull the conversion rate json isn't issuing keys to this project anymore? I'm not horribly familiar with how it works, but I know that if data is not being pulled from the request, the exception (i.e. the error) will always be thrown.

myungyunC:
If this is still an open issue (which it seems to be) @goldtomato and I will take it!

We just need to make sure basic conversions work correct? Showing conversion rates and actually converting values to the desired currency?

abhikot:
I will work on this issue! I'll investigate the api.fixer.io problem that ChristopherDaw mentioned above and potentially re-work the code and test locally.

myungyunC:
@abhikot Hi! Goldtomato and I have already started working on it, and I would prefer to not just give up our work, so if it's  alright we  will continue to work on it :) we will make a pull request soon.

abhikot:
@myungyunC Hello! I've also began working on it. Perhaps we can both try to fix it and collaborate soon to most effectively fix the issue and have our pull request accepted.

PR #952, opened 2019-12-01
Fixed and improved the Currency Module.

myungyunC:
Following issue #482 

The existing currency module was pulling from fixer.io, which only enables a set amount of API pulls for a free trial. The new proposed API exchangerates.io is an existing currency exchange API which uses fixer, thus enabling JARVIS to bypass the limitations.

This pull request not only fixes the module, cleans up the code to be more readable, and changes the API, but it also allows for better conversions. Previously, the module would not recognize terms such as 'won', 'yen', or 'dollar'. However, it can now recognize these and convert them appropriately.

I will update the test file after this pull request is accepted. As this is my first contribution, I would greatly appreciate any criticism, advice, or tips in general!

Output for query "yen to peso": 
_{"text": "Conversion Rate from JPY to MXN: 0.1781"}_

Output for query "100 usd to won":
_{"text": "100 USD is equal to 117993.99 KRW"}_

- API changed from fixer.io to exchangerates.io
- Added comments to improve readability
- Enabled queries with regular currency terms. (i.e. "dollar" instead of "USD")

**Fixes**: #

### DIFF
--- a/modules/src/currency.py
+++ b/modules/src/currency.py
@@ -1,28 +1,85 @@
+"""
+Currency Module
+
+This module can take in queries to either return a set
+conversion rate from one currency to another, or it will
+return the appropriate converted result.
+
+API used: https://exchangeratesapi.io/
+"""
 import requests
 
 from templates.text import TextTemplate
 
 
 def process(input, entities):
+    # Official Currency Terms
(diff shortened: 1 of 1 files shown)


myungyunC:
@swapagarwal

Issue #364, opened 2017-12-19
"Is there anything else?" handles improperly


cyrokamogawa:
When JARVIS asks: "Is there anything else?"
I enter "yes," but JARVIS does not understand what I mean.
I think I can resolve this issue.

yoodanxd:
I would like to work on this issue if it hasn't been completed yet.

PR #956, opened 2019-12-02
Added Quick Replies to the Hello Module for More Interaction

yoodan:

This pull request is for a bug fix/modification on an existing module seen here: #364. All of the responses from the hello.py module are handled the same way by the user, but some seem to assume a response from the user. This adds more options for the user so that they aren't completely lost in what to do afterward.

- modules/src/hello.py is the only source code file updated in this pull request. It is a simple change where I added quick_replies so that the user has more options in what to do after JARVIS gives them a response to a "hello" query. 
- I believe adding quick_replies is a great way, in general, to gauge more responses from the user. I would like to add more quick_replies to other modules that do not have them. I would also like to suggest removing the "is there anything else?" response from the hello.py and perhaps put it in a different module. This response as a whole doesn't quite fit in with the other hello.py responses.

**Fixes**: #


yoodan:
I accidentally added commits for another module I worked on a few days ago. Added some commits reverting those files so this pull request contains _only_ the changes I made for the Hello Module.

Issue #341, opened 2017-11-15
how about add a love expression module on Jarvis?


ghdud4006:
ex)
user :  i love you 
jarvis : "I love you too, sir"

user : i like you
jarvis : "I truly love you more than I have ever loved anyone"

user : do you love me?
jarvis : "I love you to moon and back."

skakim:
#355 is a PR related to this.

myungyunC:
I saw the already implemented love module which seems very limited. Can I take on this issue and attempt to improve and add to it?

PR #959, opened 2019-12-06
Added a Love Module

myungyunC:

Addresses Issue: #341 
This module detects if the user query has a 'love' intent and
returns an appropriate response which is either a response to
the user's love or a definition of 

#### Example output:
Input: "Do you love me?"
Output: {"text": "I truly love you more than I have ever loved anyone."}

Input: "What is love?"
Output: {"text": "Love is patient, love is kind. It does not envy, and it does not boast."}

- Added a local JSON file for love responses
- Added a love module test
- Checks if the user query has love intent.

**Fixes**: #


myungyunC:
@swapagarwal

Issue #1114, opened 2023-09-24
Requirements.txt Install Bugs


gigoiy:
## Steps taken to reproduce this bug
1. Clone `main` branch repo
2. `pip install -r requirements.txt`

## Results
Terminal returns with `Error: Incompatible version of Python`

## Expecting
Terminal returning with successful installation message

## Versions of Software
**Operating System:** Windows 11 64-bit
**VS Code:** 1.82.2
**Python:** 3.11

I'm currently working on this fix.

legendkartik45:
Is this issue still open?

PR #1115, opened 2023-09-24
Update requirements.txt to work with most recent versions of Python

gigoiy:
#### Description:
This update resolves outdated dependencies with the most recent version of Python, causing errors when installing requirements.txt

- Update requirements.txt to contain dependencies with the most recent versions

**Fixes**: #1114


gigoiy:
Didn't realize two other users are already doing this. #502

